PR #59, opened 2025-10-24
Feature/codecheck metadata form

subhanLabs:
Add CODECHECK metadata form to review workflow

- Create dedicated CODECHECK tab in workflow sidebar under Review stage
- Add form for capturing CODECHECK configuration version, publication method
- Implement manifest table for output files and descriptions
- Add paper metadata fields (title, authors with ORCIDs, reference)
- Add source material URL input
- Add CODECHECKers list with ORCID support
- Add CODECHECK certificate URL field with upload functionality
- Add summary textarea for CODECHECK completion notes
- Add repositories list for code/data links
- Add completion timestamp and identifier fields
- Integrate form with Vue.js components for dynamic interactions

Closes #9 

Closes #30
---

## Related User Stories

- C-MM1 (related): As a codechecker, I want the plugin to provide a form that generates, validates, and stores the codecheck.yml...
- P-IM1 (related): As a journal manager, I want to enter, edit, and view CODECHECK metadata entirely within the OJS editorial workflow...

_User story references from the [CHECK-PUB User Stories document](https://codecheck.org.uk/pub/)._

PR #63, opened 2025-10-30
Feature/UI retrieve reserve identifiers

### LINKED ISSUE #47: Author name in register issue with setting
Relates to #11 

- [x] the new CODECHECK Register issue includes the first author's last name for easier reference (e.g., Smith or "Smith et al.")
- [ ] #91
- [ ] the documentation around the setting makes clear that this is a potential conflict with anonymous review
---

## Related User Stories

- C-AD1 (related): As a codechecker, I want the plugin to automatically reserve a certificate identifier, create the Register issue...

### LINKED ISSUE #48: UI/ Frontend for: Retrieving and reserving of Certificate Identifer
Closely relates to #11

- [ ] create the front-end based on the following mock-ups:

Identifier already set:

Image

Identifier not set: 

Image

The UI is closely connected with #9.

- - -

- [ ] connect this Front-end with the PHP script from #11

### LINKED ISSUE #64: Certificate Identifier Fields in CODECHECK Metadata Form in Editorial Workflow
**Depends on #48, and PR #63**

After Adding the UI Fields to the Submission form, they now have to be added to the CODECHECK Metadata Form created by Subhan (#59).

## What we have:

Image

## What we want there:

Image
---

## Related User Stories

- P-IM1 (related): As a journal manager, I want to enter, edit, and view CODECHECK metadata entirely within the OJS editorial workflow...
- E-PM1 (related): As an editor, I want to see and edit CODECHECK metadata in Publication > Metadata tab...

dxL1nus:
Closes: #47, closes: #48, closes: #64;

## IMPORTANT:
This needs to be merged after https://github.com/codecheckers/ojs-codecheck/pull/58, as it utilized the Vue refactor form the branch that #58 is based on. So any changes made through #58 should also be in this PR afterwards.

- - -

Added The features to retrieve and reserve identifiers via the UI. At this point in time, this is only in the submission form, but once the other PRs are merged, I will also implement this in our editorial workflow pages.

- - -

Click `Reserve Identifier`:

grafik

Get Alert once Identifier has been reserved and new issue was created inside: https://github.com/codecheckers/testing-dev-register/issues

grafik

When Identifier is reserved, the UI now looks like this:

grafik

When the identifier gets deleted, it now looks like this:

grafik

It it correctly displayed in the created YAML file:
grafik


CodecheckPlugin.php
```diff
@@ -1,178 +1,221 @@
...
+                        $genres = $genreDao->getByContextId($context->getId());
+                        
+                        $codecheckGenres = [];
+                        while ($genre = $genres->next()) {
+                            $name = $genre->getLocalizedName();
+                            if (is_array($name)) {
+                                $name = $name['en'] ?? $name[array_key_first($name)] ?? '';
+                            }
+                            
+                            if (str_contains($name, 'CODECHECK') || str_contains($name, 'codecheck')) {
```
nuest: I think it would be more readable here to transform $name to lowercase and then make a contains check only once, see `strtolower`.

nuest:
@dxL1nus will update this PR to _not_ include the certificate identifier in the submission form.

classes/RetrieveReserveIdentifiers/CertificateIdentifierList.php
```diff
@@ -0,0 +1,141 @@
...
+// get the certificate ID from the issue description
+function getRawIdentifier(string $title): string
+{
+    $title = strtolower($title); // convert whole title to lowercase
+
+    //$title = "Arabsheibani, Winter, Tomko | 2025-026/2025-029";
+
+    if (strpos($title, '|') !== false) {
+        // find the last "|"
+        $seperator = strrpos($title, '|');
```
nuest: this is a typical case where a regex works also nicely, and IMO is better than manually adding the index in the string. here is what ChatGPT suggests for "text after the last occurence of |":

```php
preg_match('/[^|]+$/', $string, $matches);
$result = $matches[0] ?? '';
```

nuest (changes requested):
I only gave some feedback because this is probably best to properly review once we have merged #59 - agreed, @dxL1nus ?

classes/RetrieveReserveIdentifiers/CertificateIdentifierList.php
```diff
@@ -0,0 +1,141 @@
...
+        // move one position forwards (so we get character after '|')
+        $seperator++;
+
+        // Find where the next line break occurs after "certificate"
+        $rawIdentifier = substr($title, $seperator);
+        // remove white spaces
+        $rawIdentifier = preg_replace('/[\s]+/', '', $rawIdentifier);
+    }
```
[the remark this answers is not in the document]
dxL1nus: That is correct. I have changed it to initialize `$rawIdentifier` as an empty string before the if and then late after `getRawIdentifier()` is called I check if the return string is empty -> if so I skip this iteration, as this function is called in a loop over all issues.

```php
// get the certificate ID from the issue description
function getRawIdentifier(string $title): string
{
    // convert whole title to lowercase
    $title = strtolower($title);

    $rawIdentifier = '';

    if (strpos($title, '|') !== false) {
        // do the if clause
    }

    return $rawIdentifier;
}
```

and then when `getRawIdentifier()` is called in line 50:

```php
foreach ($apiParser->getIssues() as $issue) {
    // raw identifier (can still have ranges of identifiers);
    $rawIdentifier = getRawIdentifier($issue['title']);
    
    // check if the identifier is empty and not set -> if so skip this identifier and move onto the next issue
    if($rawIdentifier == '') {
        continue;
    }

    // append to all identifiers in new Register
    $newCertificateIdentifierList->appendToCertificateIdList($rawIdentifier);
}
```

I check if the `$rawIdentifier` is empty and if so I skip the iteration.
Is this a good solution @nuest?

classes/RetrieveReserveIdentifiers/CertificateIdentifierList.php
dxL1nus: That is a nice shorter way to do it. It now looks like this:

```php
$rawIdentifier = '';

if (strpos($title, '|') !== false) {
    // split the title into sub-strings at separator letter: '|'
    // store those sub-strings in the $matches array
    preg_match('/[^|]+$/', $title, $matches);

    // $matches[0] is the last sub-string so here the Certificate Identifier
    // when no '|' would exist then it would be the whole string -> but this case is excluded because of the if statement in line 17
    $rawIdentifier = preg_replace('/[\s]+/', '', $matches[0] ?? '');
}

return $rawIdentifier;
```

CodecheckPlugin.php
```diff
@@ -1,179 +1,176 @@
...
-                    if ($hasCodecheckOptions) {
-                        error_log("CODECHECK: Options already exist in form, skipping");
-                        return false;
-                    }
-                    
-                    // Create genres if needed (only once)
-                    $this->createCodecheckGenres();
+            Hook::add('Submission::validate', $this->saveWizardFieldsFromRequest(...));
+            // Add hook for Ajax API calls
+            Hook::add('LoadHandler', [$this, 'setupAPIHandler']);
```
dxL1nus: use the Hook `'Dispatch::dispatch'` here instead as mentioned [here](https://github.com/pkp/pkp-lib/discussions/11991#discussioncomment-14881792) and shown in [this example](https://github.com/lepidus/customQuestions/blob/main/CustomQuestionsPlugin.php#L106-L126).

CodecheckPlugin.php
```diff
@@ -1,179 +1,176 @@
...
+            // Add hook for the Template Manager
+            Hook::add('TemplateManager::display', $this->callbackTemplateManagerDisplay(...));
+            
+            $metadataHandler = new CodecheckMetadataHandler($this);
+            Hook::add('LoadHandler', function($hookName, $args) use ($metadataHandler) {
```
dxL1nus: merge Subhans and Linus APIHandler approaches to use one singular API handler

classes/RetrieveReserveIdentifiers/CertificateIdentifierList.php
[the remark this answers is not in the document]
dxL1nus: It now initializes `$rawIdentifier` to `null` and later checks if it is empty as can be seen in commit [c1f24a3](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/c1f24a31f6534f58d700907ee7fe783d5e0ee43d).

```php
// get the certificate ID from the issue description
function getRawIdentifier(string $title): ?string
{
    // convert whole title to lowercase
    $title = strtolower($title);

    $rawIdentifier = null;

    if (strpos($title, '|') !== false) {
        // split the title into sub-strings at separator letter: '|'
        // store those sub-strings in the $matches array
        preg_match('/[^|]+$/', $title, $matches);

        // $matches[0] is the last sub-string so here the Certificate Identifier
        // when no '|' would exist then it would be the whole string -> but this case is excluded because of the if statement in line 17
        $rawIdentifier = preg_replace('/[\s]+/', '', $matches[0] ?? '');
    }

    return $rawIdentifier;
}
```

check if it is empty:

```php
foreach ($apiParser->getIssues() as $issue) {
    // raw identifier (can still have ranges of identifiers);
    $rawIdentifier = getRawIdentifier($issue['title']);
    
    // check if the identifier is empty (either empty string or null) and not set
    // -> if so skip this identifier and move onto the next issue
    if(empty($rawIdentifier)) {
        continue;
    }

    // append to all identifiers in new Register
    $newCertificateIdentifierList->appendToCertificateIdList($rawIdentifier);
}
```

classes/RetrieveReserveIdentifiers/JsonApiCaller.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function fetch()
+    {
+        // Fetch JSON from API
+        $response = file_get_contents($this->url);
+
+        // throw error if no data was fetched from API
+        if ($response === FALSE) {
+            throw new ApiFetchException("Error fetching the API data");
```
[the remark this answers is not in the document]
dxL1nus: I now catch it in a way, that the codecheck API will send a `success = false` together with the Exception Message, if an Exception happens. This will definitely also need to be Unit tested.

See commit: [06eb9a7](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/06eb9a76c9845c755e1d23e4f3df91cd102aa484)

classes/CodecheckRegister/CodecheckVenueTypes.php
```diff
@@ -0,0 +1,34 @@
...
+namespace APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers;
+
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\UniqueArray;
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\JsonApiCaller;
+
+class CodecheckVenueTypes
+{
+    private UniqueArray $uniqueArray;
+
+    function __construct()
```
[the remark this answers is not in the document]
dxL1nus: The API Handler now outputs `success = false` together with the Exception Message (see commit: [06eb9a7](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/06eb9a76c9845c755e1d23e4f3df91cd102aa484))

classes/CodecheckRegister/CodecheckVenueNames.php
```diff
@@ -0,0 +1,43 @@
...
+
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\UniqueArray;
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\CodecheckRegisterGithubIssuesApiParser;
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\CodecheckVenueTypes;
+
+class CodecheckVenueNames
+{
+    private UniqueArray $uniqueArray;
+
+    function __construct()
```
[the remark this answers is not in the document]
dxL1nus: The API Handler now outputs `success = false` together with the Exception Message (see commit: [06eb9a7](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/06eb9a76c9845c755e1d23e4f3df91cd102aa484))

CodecheckPlugin.php
dxL1nus: See commit [7b57cf9](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/7b57cf93fccc3f10fdb6954c34d17262ef525e03)

We are now have our very own API endpoint called `api/v1/codecheck/{ourRoute}` with our own API handler that behaves similarly to the old APIHandler in OJS 3.4.x. Finally I made it work thank's to the `Dispatcher::dispatch` Hook

CodecheckPlugin.php
dxL1nus: Merged the `CodecheckMetadataHandler` by @Subhanaliweb and my `CodecheckApiHandler` and removed Subhans `submission\CodecheckApiHandler` entirely.
Now everything that is only related to the metadataHandler like `buildYaml()` or `getAuthors()` is inside the **Metadata Handler** and every function that returns JSON Data for the API is inside the API Handler.
I also added Subhans metadata routes to the API Handler and removed this `'LoadHandler'` Hook.

Now the whole metadata functions utilize `api/v1/codecheck/{metadataRoute}`.

See commit [db508e8](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/db508e889fd786434b7d4f3047993b779a7d380a)

dxL1nus: Our API Handler now correctly checks for the correct CSRF Token when an API route is requested with methods different to 'GET' e.g. 'POST'.

See commit [329a14f](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/329a14fe3fe80d403016ea11fd508fd8ade9eca2)

classes/RetrieveReserveIdentifiers/CertificateIdentifierList.php
```diff
@@ -0,0 +1,141 @@
+<?php
+
+namespace APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers;
+
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\UniqueArray;
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\CodecheckRegisterGithubIssuesApiParser;
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\CertificateIdentifier;
+
+// get the certificate ID from the issue description
+function getRawIdentifier(string $title): string
```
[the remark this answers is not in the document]
dxL1nus: I opened a new issue for this (#70) that I will now work on

classes/CodecheckRegister/CertificateIdentifierList.php
```diff
@@ -0,0 +1,141 @@
...
+        // Find where the next line break occurs after "certificate"
+        $rawIdentifier = substr($title, $seperator);
+        // remove white spaces
+        $rawIdentifier = preg_replace('/[\s]+/', '', $rawIdentifier);
+    }
+
+    return $rawIdentifier;
+}
+
+class CertificateIdentifierList
```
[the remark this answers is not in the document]
dxL1nus: See Issue #70

dxL1nus:
I now merged main into the UI branch @nuest

classes/RetrieveReserveIdentifiers/CertificateIdentifierList.php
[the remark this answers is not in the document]
dxL1nus: The Unit Tests now work, but they will have their own PR

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+            Role::ROLE_ID_MANAGER,
+            Role::ROLE_ID_SUB_EDITOR,
+            Role::ROLE_ID_ASSISTANT,
+            Role::ROLE_ID_AUTHOR
+        ];
+
+        $this->endpoints = [
+            'GET' => [
+                [
+                    'route' => 'getVenueData',
```
nuest: The route should not contain a verb - that is what the `GET` in the HTTP operation is for. Also, the "Data" is redundant, we always want to retrieve data via an API, or if we want something else (more specific), we can add sub-resources (e.g., venue/location or venue/image).

So please rename this route to `venue`.

The function name can stay the same, there the verb and the detail about the value of the returned result make sense.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+            'size' => $file['size']
+        ], 200);
+    }
+
+    /**
+     * Download a file from the CODECHECK manifest
+     * 
+     * @return void
+     */
+    public function downloadFile(): void
```
nuest, line 508: Please clarify what this upload and download is used for. The manifest is primarily _metadata_ and IIRC we are currently not implementing the case that a manifest file is actually uploaded to OJS. I understand this may be useful in the future where an author may be able to privately share manifest files this way, but first we handle the case of a check based on an external code/data repository.

I'm not saying we need to take these functions out, but they are probably dormant for now.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+                    'handler' => [$this, 'getMetadata'],
+                    'roles' => $this->roles,
+                ],
+                [
+                    'route' => 'download',
+                    'handler' => [$this, 'downloadFile'],
+                    'roles' => $this->roles,
+                ],
+                [
+                    'route' => 'yaml',
```
nuest, line 69: Good route name!

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+use APP\plugins\generic\codecheck\classes\RetrieveReserveIdentifiers\CodecheckVenue;
+use APP\plugins\generic\codecheck\classes\Workflow\CodecheckMetadataHandler;
+
+use APP\facades\Repo;
+use Illuminate\Support\Facades\DB;
+
+class CodecheckApiHandler
+{
+    private JsonResponse $response;
+    private array $roles;
```
nuest, line 26: For my understanding: where do we assign the required roles? Or is this handled by OJS core?

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+                ],
+                [
+                    'route' => 'yaml',
+                    'handler' => [$this, 'generateYaml'],
+                    'roles' => $this->roles,
+                ],
+            ],
+            'POST' => [
+                [
+                    'route' => 'reserveIdentifier',
```
nuest: Similarly as above, we avoid verbs in routes, so this can simply be `POST identifier` and then it is not a concern of the caller of the API what kind of "creation" this means, so in this case an external system is used to "reserve something", but it could also be a local random string generator. From the API perspective, we just say "call this endpoint when you want to create a new identifier".

CodecheckPlugin.php
```diff
@@ -95,6 +62,29 @@ public function register($category, $path, $mainContextId = null): bool
...
+    {
+        $request = $args[0];
+        $router = $request->getRouter();
+
+        if (!($router instanceof \PKP\core\APIRouter)) {
+            return;
+        }
+
+        if (str_contains($request->getRequestPath(), 'api/v1/codecheck')) {
+            error_log("[CODECHECK Plugin] Instanciating the CODECHECK APIHandler");
```
nuest, line 75: Why is this an "error" log? Please use a different log function here. This also applies to several other uses of this log function. Errors are unexpected or problematic cases, not informative messages.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+     */
+    private function serveRequest(): void
+    {
+        // get the request Method like POST or GET
+        $method = $this->request->getRequestMethod();
+
+        error_log("Method: " . $method);
+
+        foreach ($this->endpoints[$method] as $endpoint) {
+            if($endpoint['route'] == $this->route) {
```
nuest: Minor style comment for readability: put the `$this->` at the same side for both calls, so `if($this->rout ==...`.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+            $certificateIdentifierList->sortDesc();
+
+            // create the new unique Identifier
+            $new_identifier = CertificateIdentifier::newUniqueIdentifier($certificateIdentifierList);
+
+            // create the CODECHECK Venue with the selected type and name
+            $codecheckVenue = new CodecheckVenue();
+
+            $codecheckVenue->setVenueType($venueType);
+            $codecheckVenue->setVenueName($venueName);
```
nuest: When you use a constructor with parameters, these three lines become one without loosing clarity.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+
+            // Add the new issue to the CODECHECK GtiHub Register
+            try {
+                $issueGithubUrl = $apiParser->addIssue($new_identifier, $codecheckVenue->getVenueType(), $codecheckVenue->getVenueName(), $authorString);
```
nuest: The name here is a bit confusing: why can I tell a "parser" to to something like "add" ?

Suggest to rename the object to `githubRegister` and the class as well to `CodecheckGitHubRegister` so that you make clear that this class offers all features for querying and managing the register.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+     * This function saves all the CODECHECK Metadata to the Database
+     * 
+     * @return void
+     */
+    public function saveMetadata(): void
+    {
+        // get submissionId
+        $submissionId = $this->codecheckMetadataHandler->getSubmissionId();
+
+        error_log("[CODECHECK Api] saveMetadata called for submission: $submissionId");
```
nuest: I like the prefix for the log statement!

API is an abbreviation, so please use "API".

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+
+        if ($exists) {
+            DB::table('codecheck_metadata')
+                ->where('submission_id', $submissionId)
+                ->update($metadataData);
+            error_log("[CODECHECK Api] Updated existing record");
+        } else {
+            $metadataData['created_at'] = date('Y-m-d H:i:s');
+            DB::table('codecheck_metadata')->insert($metadataData);
+            error_log("[CODECHECK Metadata] Created new record");
```
nuest, line 404: For clarity, please add the submission ID to the log statements here (and a few lines above). "Record" can also be confusing, so either add "... record in the CODECHECK metadata table", or maybe "Inserted/Updated codecheck metadata for submission XYZ".

Imagine two authors submitting information almost at the same time, and something goes wrong during one of them - we want to identify the log statements of relevance then.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+        }
+
+        $this->response->response([
+            'success' => true,
+            'message' => 'CODECHECK metadata saved successfully'
```
nuest: For a possible validation on the client side, I suggest to also return the submission ID in the response and the certificate identifier if it is already set.

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+        $submissionId = $this->codecheckMetadataHandler->getSubmissionId();
+
+        error_log("[CODECHECK Metadata] generateYaml called for submission: $submissionId");
+        
+        $submission = Repo::submission()->get($submissionId);
+        
+        if (!$submission) {
+            $this->response->response([
+                'success' => false,
+                'error' => 'Submission not found'
```
nuest: For better debugging and possible validation: include the submission ID that was not found in the response:

```php
'error' => 'Submission $submissionId not found'
'subimssionID' => $submissionId
```

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+        $publication = $submission->getCurrentPublication();
+        
+        $metadata = DB::table('codecheck_metadata')
+            ->where('submission_id', $submissionId)
+            ->first();
+
+        if (!$metadata) {
+            $this->response->response([
+                'success' => false,
+                'error' => 'No CODECHECK metadata found'
```
nuest: Include submission ID in the error response (see above).

api/v1/CodecheckApiHandler.php
```diff
@@ -0,0 +1,582 @@
...
+            ], 404);
+            return;
+        }
+
+        $yaml = $this->codecheckMetadataHandler->buildYaml($publication, $metadata);
+
+        $this->response->response([
+            'success' => false,
+            'yaml' => $yaml,
+            'filename' => 'codecheck.yml'
```
nuest, line 592: Does this mean the response is a file, or a JSON object where the YAML is embedded?

classes/Exceptions/NoMatchingIssuesFoundException.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace APP\plugins\generic\codecheck\classes\Exceptions;
+
+class NoMatchingIssuesFoundException extends \Exception
+{
+    public function __construct(
+        string $message = "No more issues available from GitHub API",
```
nuest: The message does not fit the class name - is this about "matching" or "no more" - or both?

I think the case of "no more issues" = end of a list should not be an error case but handled in the regular flow of the code.

classes/CodecheckRegister/CertificateIdentifierList.php
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * Get the latest/ newest Certificate Identifier
+     * 
+     * @return CertificateIdentifier Returns the newest Certificate Identifier
+     */
+    public function getNewestIdentifier(): CertificateIdentifier
+    {
+        $this->sortDesc();
+        // get first element of sort descending -> newest element
+        return $this->uniqueArray->at(0);
```
nuest, line 188: `sortDesc` already does a `uniqueArray`.

classes/RetrieveReserveIdentifiers/CodecheckRegisterGithubIssuesApiParser.php
```diff
@@ -0,0 +1,155 @@
...
+     */
+    public function fetchIssues(): void
+    {
+        $issuePage = 1;
+        $issuesToFetchPerPage = 20;
+        $fetchedMatchingIssue = false;
+
+        do {
+            try {
+                $allissues = $this->client->api('issue')->all('codecheckers', 'testing-dev-register', [
```
nuest: Hardcoded to testing-dev-register here, probably not what we want!

classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php
```diff
@@ -0,0 +1,155 @@
...
+                    'per_page'  => $issuesToFetchPerPage, // issues that will be fetched per page
+                    'page'      => $issuePage,
+                ]);
+            } catch (\Throwable $e) {
+                throw new ApiFetchException("Failed fetching the GitHub Issues\n" . $e->getMessage());
+            }
+
+            // stop looping if no more issues exist and we haven't yet found a matching issue
+            if (empty($allissues) && empty($this->issue)) {
+                throw new NoMatchingIssuesFoundException("There was no Issue found with a '|' inside the GitHub Codecheck Register.");
```
nuest: AFAICS there is not filtering here for the pipe character - please double check the exception message.

classes/RetrieveReserveIdentifiers/CodecheckRegisterGithubIssuesApiParser.php
```diff
@@ -0,0 +1,155 @@
...
+        } while (!$fetchedMatchingIssue);
+    }
+
+    /**
+     * Fetches a Issue Labels from the CODECHECK GitHub Register
+     */
+    public function fetchLabels(): void
+    {
+        try {
+            $fetchedLabels = $this->client->api('issue')->labels()->all('codecheckers', 'testing-dev-register');
```
nuest: Hardcoded dev register.

classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php
```diff
@@ -0,0 +1,155 @@
...
+        string $codecheckVenueType,
+        string $codecheckVenueName,
+        string $authorString,
+    ): string {
+        $token = $_ENV['CODECHECK_REGISTER_GITHUB_TOKEN'];
+
+        $this->client->authenticate($token, null, Client::AUTH_ACCESS_TOKEN);
+
+        $repositoryOwner = 'codecheckers';
+        $repositoryName = 'testing-dev-register';
```
nuest: Hardcoded dev register.

Maybe for now you make that a class variable/constant and then we turn it into a plugin configuration setting?

classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php
```diff
@@ -0,0 +1,155 @@
...
+        $authorString = empty($authorString) ? 'New CODECHECK' : $authorString;
+        $issueTitle = $authorString . ' | ' . $certificateIdentifier->toStr();
+        $issueBody = '';
```
nuest: For now, please put the journal name and the submission ID into the body, so that we have some context information that is not sensitive.

classes/CodecheckRegister/CodecheckGithubRegisterApiClient.php
```diff
@@ -0,0 +1,155 @@
...
+        $token = $_ENV['CODECHECK_REGISTER_GITHUB_TOKEN'];
+
+        $this->client->authenticate($token, null, Client::AUTH_ACCESS_TOKEN);
+
+        $repositoryOwner = 'codecheckers';
+        $repositoryName = 'testing-dev-register';
+        $authorString = empty($authorString) ? 'New CODECHECK' : $authorString;
+        $issueTitle = $authorString . ' | ' . $certificateIdentifier->toStr();
+        $issueBody = '';
+        $labelStrings = ['id assigned'];
```
nuest, line 125: Nothing to do now, but take note for later: https://github.com/codecheckers/ojs-codecheck/issues/81

classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php
```diff
@@ -0,0 +1,155 @@
...
+                $repositoryOwner,
+                $repositoryName,
+                [
+                    'title' => $issueTitle,
+                    'body'  => $issueBody,
+                    'labels' => $labelStrings
+                ]
+            );
+        } catch (\Throwable $e) {
+            throw new ApiCreateException("Error while adding the new GitHub issue with the new Certificate Identifier\n" . $e->getMessage());
```
nuest: Add the certificate identifier to the exception text.

classes/CodecheckRegister/CodecheckVenueNames.php
```diff
@@ -0,0 +1,60 @@
...
+        // Initialize unique Array
+        $this->uniqueArray = new UniqueArray();
+
+        // Intialize API caller
+        $jsonApiCaller = new JsonApiCaller("https://codecheck.org.uk/register/venues/index.json");
+        // fetch CODECHECK Type data
+        try {
+            $jsonApiCaller->fetch();
+        } catch (ApiFetchException $e) {
+            // TODO: Implement that the user gets notified, that the fetching of the Labels didn't work
```
nuest, line 28: Add an issue for open tasks, link to the issue from the TODO comment.

classes/CodecheckRegister/CodecheckVenueNames.php
```diff
@@ -0,0 +1,60 @@
...
+            error_log($e);
+            throw $e;
+            return;
+        }
+        // get json Data from API Caller
+        $data = $jsonApiCaller->getData();
+
+        // find all venue Types
+        // TODO: Remove this once the actualy Codecheck API contains the labels/ Venue Names to fetch
```
nuest, line 42: https://codecheck.org.uk/register/venues/index.json returns (excerpt):

```json
[
  {
    "Venue type": "journal",
    "Venue name": "GigaScience",
    "No. of codechecks": 2,
    "Issue label": "journal"
  },
  {
    "Venue type": "community",
    "Venue name": "CODECHECK",
    "No. of codechecks": 11,
    "Issue label": "community"
  },
  {
    "Venue type": "community",
    "Venue name": "Preprint",
    "No. of codechecks": 9,
    "Issue label": "preprint"
  },
  {
    "Venue type": "community",
    "Venue name": "In press",
    "No. of codechecks": 1,
    "Issue label": "journal"
  },
  {
    "Venue type": "journal",
    "Venue name": "Journal of Geographical Systems",
    "No. of codechecks": 2,
    "Issue label": "journal"
  },
  {
    "Venue type": "conference",
    "Venue name": "AGILE Conference on Geographic Information Science",
    "No. of codechecks": 59,
    "Issue label": "conference"
  },

[...]
```

What is still missing?

classes/CodecheckRegister/JsonApiCaller.php
```diff
@@ -0,0 +1,48 @@
...
+     * This function fetches all the data from the given URL
+     */
+    public function fetch()
+    {
+        // Fetch JSON from API
+        $response = file_get_contents($this->url);
+
+        // throw error if no data was fetched from API
+        if ($response === FALSE) {
+            throw new ApiFetchException("Error fetching the Codecheck API data");
```
nuest: Add the url to the exception, either by adding specific fields (it is your own exception class) or by adding it to the message text.

Does the response not contain any information when there is an error? Maybe an error message or an HTTP status code?

classes/DataStructures/UniqueArray.php
```diff
@@ -0,0 +1,104 @@
...
+     * This function adds a new element to the `UniqueArray`
+     * 
+     * @param mixed $element The new Element that is being added to the `UniqueArray` (of any Type)
+     */
+    public function add($element): void
+    {
+        if(!$this->contains($element)) {
+            $this->array[] = $element;
+        }
+        $this->array = array_values(array_unique($this->array, SORT_REGULAR));
```
nuest: This is double effort: you only add an element when it does not exist, and then you call `array_unique` again. Stick to one way.

classes/DataStructures/UniqueArray.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * This function gets an element inside the `UniqueArray` at a sepecific index
+     * 
+     * @return mixed This function either returns an element of any Type at the index from the `UniqueArray` or `null` if no element was found at that index
+     */
+    public function at(int $index): mixed
+    {
+        return $this->array[$index] ?? null;
+    }
```
nuest, line 55: Are these functions actually used in our code? I understand these are needed for a "complete" implementation of an array, but they also create code that needs to be maintained. I'd like to stick to the functions that we really use for now.

classes/DataStructures/UniqueArray.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * This function checks if a specific Element exists inside the `UniqueArray`
+     * 
+     * @return bool Returns `true` if the element exists inside the `UniqueArray` and `false` if otherwise
+     */
+    public function contains($searchElement): bool
+    {
+        foreach ($this->array as $arrayElement) {
```
nuest: Prefer built-in functions over own loop-based implementations, see https://www.php.net/manual/en/function.in-array.php

Rewrite the contains functions to use `in_array`, which I would trust to be fast and optimised.

locale/en/locale.po
```diff
@@ -53,6 +53,12 @@ msgstr "Data Repository URLs"
 msgid "plugins.generic.codecheck.dataRepository.description"
 msgstr "Link(s) to your data repository (Zenodo, OSF, etc.) - optional"
 
+msgid "plugins.generic.codecheck.dataSoftwareAvail"
```
nuest: Don't try to save a few characters here, stay readable > `..Availability`

locale/en/locale.po
```diff
@@ -53,6 +53,12 @@ msgstr "Data Repository URLs"
...
+msgstr "Data and Software Availability"
+
+msgid "plugins.generic.codecheck.dataSoftwareAvail.description"
+msgstr "Describe how your data and code are available..."
```
nuest: Please change to:

> Describe where your data, code, and documentation is available and under what conditions, e.g., copyright, licenses, sensitivity limitations, or access procedures for protected data. The information should be concise and ideally comprises (a) for code, a minimal description of the workflow environment and main tools including versions, and (b) for data, persistent links to repositories for code and data using Digital Object Identifiers (DOI), but may also be plain URLs with last access dates or even a description of data access steps. Remove links for anonymity during peer review (“xxx”), or share anonymized links if your repository supports them. Used data, software and (third-party) tools should be cited.

I know this is a long text, but I want to see how that looks in the submission form. And when we start talking to journals, we need to find out how to link this to existing author guidelines on providing this kind of information.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
+<template>
+    <div class="certificate-identifier-section">
+        <div class="certificate-identifier-input-wrapper">
+            <input
+                type="text"
+                v-model="identifier"
+                placeholder="ID - e.g.: 2025-001"
```
nuest: Placeholder needs to be localised.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+                <option v-for="type in venueTypes" :key="type" :value="type">
+                {{ type }}
+                </option>
+            </select>
+            <select
+                v-model="venueName"
+                class="certificate-identifier-select certificate-identifier-venue-names"
+                :disabled="isIdentifierReserved"
+            >
+                <option disabled value="default">Venue Name</option>
```
nuest: Default (disabled) option needs to be localised.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+                <option v-for="name in venueNames" :key="name" :value="name">
+                {{ name }}
+                </option>
+            </select>
+        </div>
+
+        <div v-if="issueUrl" class="certificate-identifier-link-wrapper">
+            <a :href="issueUrl" target="_blank">
+                View GitHub Issue
```
nuest: Add localisation to link text.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+                @click="reserveIdentifier"
+            >
+                Reserve Identifier
+            </button>
+            <button
+                type="button"
+                class="certificate-identifier-button bg-red"
+                @click="removeIdentifier"
+            >
+                Remove Identifier
```
nuest: Add localisation to button texts.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+}
+
+async function reserveIdentifier() {
+    if (venueType.value === 'default' || venueName.value === 'default') {
+        alert('Please select both a Venue Type and a Venue Name.');
+        return;
+    }
+
+    let codecheckApiUrl = pkp.context.apiBaseUrl.replace(/\/api\/v1\/?$/, '');
+    codecheckApiUrl += '/codecheck_api';
```
nuest: The same code is in lines 89 and 90 > define the API endpoint only once.

Also, in the regex for matching requests in the API handler https://github.com/codecheckers/ojs-codecheck/pull/63/changes#diff-b807d8c7cd429d2b42ac67336f8803632abdc50a1675f4109979b3850760e819 the API URL is only /codecheck/.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+
+    try {
+        const response = await fetch(`${codecheckApiUrl}/reserveIdentifier`, {
```
nuest: Rename endpoint as described in handler class.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+            venueType: venueType.value,
+            venueName: venueName.value,
+            }),
+        });
+        const data = await response.json();
+
+        if (data.success) {
+            identifier.value = data.identifier;
+            issueUrl.value = data.issueUrl;
+            alert(`Added new issue with identifier: ${data.identifier}`);
```
nuest: This is fine for now, but please create an issue to add a nicer user notification than a plain JS alert.

resources/js/Components/CodecheckCertificateIdentifier.vue
```diff
@@ -0,0 +1,252 @@
...
+            console.log('Added new issue with identifier: ', data.identifier, data.issueUrl);
+        } else {
+            console.error('Error:', data.error);
+        }
+    } catch (error) {
+        console.error('Request failed:', error);
```
nuest: Same here: fine for now, but create issues to add proper user notifications.

resources/js/Components/CodecheckDataAndSoftwareAvailability.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+  <div class="codecheck-data-and-software-availability">
+    <textarea
+      :value="dataSoftwareAvail"
+      @input="onInput"
+      placeholder="Describe how your data and code are available..."
```
nuest: Localise placeholder.

nuest (changes requested):
Re file structure and naming: the folder "classes/RetrieveReserveIdentifiers/" is not consistent with the code in it, there is also code about venue types and names; suggest to rename, maybe to "classes/Register" ?

See this file for examples on how to use localised strings in a `.vue` file: https://github.com/codecheckers/ojs-codecheck/blob/main/resources/js/Components/CodecheckMetadataForm.vue

dxL1nus:
> Re file structure and naming: the folder "classes/RetrieveReserveIdentifiers/" is not consistent with the code in it, there is also code about venue types and names; suggest to rename, maybe to "classes/Register" ?

Then I would call it `classes/CodecheckRegister`, because `register` in of itsself is a bit confusing, as it could e.g. also mean that there are special scripts inside that help the plugin get registered in OJS or something else

dxL1nus:
> See this file for examples on how to use localised strings in a `.vue` file: https://github.com/codecheckers/ojs-codecheck/blob/main/resources/js/Components/CodecheckMetadataForm.vue

I am aware on how to use localised strings, have I forgotten to do that soemwhere in the code? If so, then I am sorry and I will fix it

api/v1/CodecheckApiHandler.php
dxL1nus: okay makes sense

api/v1/CodecheckApiHandler.php
dxL1nus: This has nothing to do with getting contents from a `codecheck.yml` file from an external repository. This is code created by @Subhanaliweb originally in the `CodecheckMetadataHandler`, which I just moved to the `CodecheckApiHandler` since it is a function, that handles an API route. You would have to ask him, which purpose this function has exactly

api/v1/CodecheckApiHandler.php
dxL1nus: So far the roles are in this array:

```php
$this->roles = [
    Role::ROLE_ID_MANAGER,
    Role::ROLE_ID_SUB_EDITOR,
    Role::ROLE_ID_ASSISTANT,
    Role::ROLE_ID_AUTHOR
];
```

and they are assigned to each route. I think it would also be possible to assign different Roles to different routes, if that is desirable and if we need it.

The roles are then checked on validity when a endpoint is calles in the authorize method:

```php
// Check if the user that accesses this resource has at least one valid Role and if user exists
$user = $this->request->getUser() ?? null;
$contextId = $this->request->getContext()->getId();

if(!($user && $user->hasRole($this->roles, $contextId))) {
    $this->response->response([
        'success'   => false,
        'error'     => "User has no assigned Role or doesn't have the right roles assigned to access this resource"
    ], 400);
    return;
}
```

api/v1/CodecheckApiHandler.php
dxL1nus: okay I changed it to just be `identifier`

CodecheckPlugin.php
dxL1nus: Well `error_log()` is just the normal default PHP log function. I don't know if another function exists, but as far as I'm concerned PHP doesn't split logs into non-errors and errors, it is all just one. So in other words, the function could well and truly also just have been called `log()` (of course not to be confused with logarithm). You can also ask @Subhanaliweb on this, as he also uses `error_log()` in his code for normal logs.

api/v1/CodecheckApiHandler.php
dxL1nus: I now use a constructor, but I kept the setter methods just in case I will need them in the future

api/v1/CodecheckApiHandler.php
dxL1nus: I believe I have called it exactly that in the past (`CodecheckGitHubRegister`), but you mean't that this would be confusing to some people, as this is not a 1 to 1 code representation of the CODECHECK GitHub Register, but more so a Handler for fetching and adding GitHub Issues. Is there a better Name to call it instead? Maybe `CodecheckGitHubRegisterHandler`?

api/v1/CodecheckApiHandler.php
dxL1nus: this is changed now to be `[CODECHECK API] ...`

api/v1/CodecheckApiHandler.php
dxL1nus: This again is code fully created by @Subhanaliweb so he would be the one to need to change that accordingly. I just again took it from the Metadata Handler and incorporated it to the API Controller

api/v1/CodecheckApiHandler.php
dxL1nus: okay

dxL1nus: Done:

```php
$this->response->response([
    'success' => true,
    'message' => 'CODECHECK metadata saved successfully',
    'submissionID' => $submissionId,
    'certificate' => $metadataData['certificate'],
], 200);
```

api/v1/CodecheckApiHandler.php
dxL1nus: This again is code produced by @Subhanaliweb originally in the `CodecheckMetadataController`. I would say, looking at the `buildYaml` function and specifically this line:

```php
// Generate YAML
$yaml = "---\n" . Yaml::dump($data, 4, 2);

return $yaml;
```

this yaml in the response is just a simple yaml file in a multi-line string, that can then again be correctly read in JS with a proper Yaml parser

classes/Exceptions/NoMatchingIssuesFoundException.php
dxL1nus: Well this is a tricky one, because essentially it is both.
The thing is we loop through all issues and this exception gets thrown if there are no more issues left to loop over. And this point can also only be reached, if no issue was found before, because the rest of the code stops the loop if at least one issue was found in an iteration. So this also means, that this exception is only reached if there was no issue, that matches the title we need (`There was no Issue found with a '|' inside the GitHub Codecheck Register.`). So in hindsight when we reach the NoMore Issues Exception it is also a NoMathingIssues Exception. Pretty confusing at first I know 😅

classes/CodecheckRegister/CertificateIdentifierList.php
dxL1nus: How is this mean't? This function is supposed to get the newest Identifier in an unsorted, but unique Array. So to get the newest, I first sort is desc to have the highest/ newest value at the first index and then I return that.

classes/RetrieveReserveIdentifiers/CodecheckRegisterGithubIssuesApiParser.php
dxL1nus: it is now more flexible:

```php
$apiParser = new CodecheckRegisterGithubIssuesApiParser('testing-dev-register');
```

```php
/**
 * Initializes a new CODECHECK GitHub Register Api Parser (initialize the GitHub Client and a new unique Array)
 * 
 * @param string $githubRegisterRepository The Repository of the GitHub Register
 */
function __construct(string $githubRegisterRepository)
{
    $this->client = new Client();
    $this->labels = new UniqueArray();
    $this->githubRegisterRepository = $githubRegisterRepository;
}
```

```php
$allissues = $this->client->api('issue')->all('codecheckers', $this->githubRegisterRepository, [
```

classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php
dxL1nus: yes there is, exactly here:

```php
do {
    try {
        $allissues = ... [
            ...
        ]);
    } catch (...) { ... }

    // stop looping if no more issues exist and we haven't yet found a matching issue
    if (empty($allissues) && empty($this->issue)) {
        throw new NoMatchingIssuesFoundException("There was no Issue found with a '|' inside the GitHub Codecheck Register.");
    }

    // !IMPORTANT! code is here:
    foreach ($allissues as $issue) {
        // here I look if the pipe is inside the issue title
        if (strpos($issue['title'], '|') !== false) {
            $this->issues[] = $issue;
            // if it is, then I set $fetchedMatchingIssue which is also the termination condition of the do while loop, so the Exception is never reached, because the loop got terminated before
            $fetchedMatchingIssue = true;
        }
    }

    $issuePage++;
} while (!$fetchedMatchingIssue); // This is the termination Condition of the loop

File: classes/RetrieveReserveIdentifiers/CodecheckRegisterGithubIssuesApiParser.php (hunk 30 above)
Comment (dxL1nus) [reply classes/RetrieveReserveIdentifiers/CodecheckRegisterGithubIssuesApiParser.php]: already fixed that

File: classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php (hunk 31 above)
Comment (dxL1nus) [reply classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php]: I fixed this together with the other fix, that the register repo is now passed as a parameter for the constructor of the `GitHubIssuesApiParser`. So this would then also be able to be changed later on in the Plugin Settings, e.g. via a setter method.

File: classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php (hunk 32 above)
Comment (dxL1nus) [reply classes/CodecheckRegister/CodecheckRegisterGithubIssuesApiParser.php]: The Body now includes the Journal Name and the Submission ID:

grafik

See [here](https://github.com/codecheckers/testing-dev-register/issues/93)

File: classes/CodecheckRegister/CodecheckVenueNames.php (hunk 35 above)
Comment (dxL1nus) [reply classes/CodecheckRegister/CodecheckVenueNames.php:28]: done here: https://github.com/codecheckers/ojs-codecheck/issues/82

File: classes/CodecheckRegister/CodecheckVenueNames.php (hunk 36 above)
Comment (dxL1nus) [reply classes/CodecheckRegister/CodecheckVenueNames.php:42]: nothing exactly. My idea was, that the codecheck API would supply all venue types and all venue names. right now for some venues the type and the name (which is retrieved through the field `"Issue label"`) is the same, so I have to instanciate all Venue Types here again and compare them with the Venue names, so Venue names are only those Issue Labels, which aren't already Venue Types. So my hope was, that for each Venue the CODECHECK API supplies a dataset, where `Venue Type` and `Issue label` are different. Do you understand what I mean @nuest?

File: classes/CodecheckRegister/JsonApiCaller.php (hunk 37 above)
Comment (dxL1nus) [reply classes/CodecheckRegister/JsonApiCaller.php]: > Does the response not contain any information when there is an error? Maybe an error message or an HTTP status code?

no because `$response` is just the return of the PHP function `file_get_contents($this->url)` which either returns the content of a file as a `string` or `false` if an error occurred (without an error code or an error message).

Comment (dxL1nus) [reply classes/CodecheckRegister/JsonApiCaller.php]: > Add the url to the exception, either by adding specific fields (it is your own exception class) or by adding it to the message text.

I just added the url to the error message, as the error message would be returned by the API as a string later, so unless we have a very specific use-case to have the extra url API field there, it is only really useful as an error message anyways.

Comment (dxL1nus) [reply classes/CodecheckRegister/JsonApiCaller.php]:
```php
// throw error if no data was fetched from API
if ($response === FALSE) {
    throw new ApiFetchException("Error fetching the Codecheck API data. URL: " . $this->url);
}
```

classes/DataStructures/UniqueArray.php
dxL1nus: I removed the lower line: `$this->array = array_values(array_unique($this->array, SORT_REGULAR));` so the double effort is now removed so the complexity should now be reduced to O(n)

classes/DataStructures/UniqueArray.php
dxL1nus: yes `at()` is used to get the newest Identifier by retrieving `uniqueArray->at(0)`.

Currently I think `remove()` is actually not used and I understand, the concern about the extra effort of maintaining it, but i already wrote Unit Tests for it as well and I definitely believe, that we will need the `remove` function for some reason or another in the future, so I am not in favor of removing it here. But it is of course your call, so can you please confirm again if you want me to remove it or not @nuest?

classes/DataStructures/UniqueArray.php
dxL1nus: I now use:

```php
/**
 * This function checks if a specific Element exists inside the `UniqueArray`
 * 
 * @return bool Returns `true` if the element exists inside the `UniqueArray` and `false` if otherwise
 */
public function contains(mixed $element): bool
{
    return in_array(
        $element,       // the element that is checked to see if it is in the array
        $this->array,   // the array
        false,           // don't use strict checking, as we sometimes want to check that objects are the same by checking their value and not their identity
    );
}
```

locale/en/locale.po
dxL1nus: Well this is again a feature of Subhan, where I just created text for it to make it look nicer. I will rename it now to `Availability`. The problem here is though, that all the other classes and ids and variable names which Subhan introduced still use the name `Avail`, so we have a name mismatch here when I rename this to `Availability`.

dxL1nus: I renamed the locale now to `...Availability`

locale/en/locale.po
dxL1nus: It now looks like this:

grafik
<br><br>
The textarea is automatically resized to always fit the whole placeholder (so in site load, window resize, and when the text is empty again and the placeholder is re-displayed).

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: As far as I'm concerned, this file is outdated anyway and not used in the code anymore, as we don't need to reserve a Certificate Identifier in the Submission Form. So it is best to delete this file entirely to avoid any further confusion.

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: this is already localized in the main metadata form vue file and because this file is deleted, this is not needed to be looked at anymore.

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: also irrelevant due to the deletion of the file

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: also irrelevant

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: yes this is still due to this file being far out of date (we are talking about probably two months out of date) and since this file is now deleted this is also irrelevant.

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: also irrelevant due to deletion of the file

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: I will change this in the main metadata form, as there I still also only use a normal alert. I will use the same PKP modal, that we use for every other alert method.

dxL1nus: This is the new modal which appears when a new identifier is reserved from now on:

grafik

dxL1nus: this is how it looks now:

```js
      .
      .
      .

      if (data.success) {
          this.metadata.certificate = data.identifier;
          this.certificateIdentifier.issueUrl = data.issueUrl;
          this.$emit('update', this.metadata.certificate);
          this.showMessage(`${this.t('plugins.generic.codecheck.identifier.reserve.success.message')}: ${data.identifier}`, 'success');
          console.log('New Certificate Identifier reserved: ', data.identifier, data.issueUrl);
      } else {
           console.error('Error while reserving the Certificate Identifier:', data.error);
      }
  } catch (error) {
      console.error('Request failed:', error);
  }
},
```

resources/js/Components/CodecheckCertificateIdentifier.vue
dxL1nus: The user is now notified like this, when the reservation of a certificate identifier fails:

grafik

dxL1nus: This is how it looks like now:

```js
      .
      .
      .
      if (data.success) {
          this.metadata.certificate = data.identifier;
          this.certificateIdentifier.issueUrl = data.issueUrl;
          this.$emit('update', this.metadata.certificate);
          this.showMessage(`${this.t('plugins.generic.codecheck.identifier.reserve.success.message')}: ${data.identifier}`, 'success');
          console.log('New Certificate Identifier reserved: ', data.identifier, data.issueUrl);
      } else {
          this.showMessage(`${this.t('plugins.generic.codecheck.identifier.reserve.fail.message')}\n${data.error}`, 'error');
          console.error('Error while reserving the Certificate Identifier:', data.error);
      }
  } catch (error) {
      this.showMessage(`${this.t('plugins.generic.codecheck.request.failed')}\n${error}`, 'error');
      console.error('Request failed:', error);
  }
},
```

resources/js/Components/CodecheckDataAndSoftwareAvailability.vue
dxL1nus: I already did that

dxL1nus:
i fixed the final requests for this PR in [3932030](https://github.com/codecheckers/ojs-codecheck/pull/63/commits/3932030e3db8c0500fc8a020f9b27eba7925460e), so maybe it can now be merged @nuest. It would be nice, if you could review it again and take a look at my answers and solutions :)

api/v1/CodecheckApiHandler.php
nuest: @dxL1nus still open comment - is there a reason against renaming the object?

api/v1/CodecheckApiHandler.php
nuest: @dxL1nus Open comment.

classes/CodecheckRegister/CertificateIdentifierList.php
nuest: @dxL1nus Please fix the duplicate "make unique".

classes/CodecheckRegister/CodecheckVenueNames.php
nuest: @dxL1nus Please clarify if this is resolved.

CodecheckPlugin.php
nuest: @dxL1nus Please try out a different log function or log level.

nuest (changes requested):
There are very open open change requests.

Also, the question about uploading and downloading files for the manifest remains - should we discuss that in our meeting?

nuest:
To avoid challenges with PRs that are based on each other, and still give me an opportunity to give feedback or review, you can also create a PR within your own fork:

https://github.com/dxL1nus/ojs-codecheck/compare/dxL1nus:feature/ui-retrieve-reserve-identifiers...dxL1nus:feature/unit-tests?expand=1

This makes the edits "on top of the other PR" readily accessible. I think I should be able to comment/review when you make me a collaborator on your repo.

classes/CodecheckRegister/CertificateIdentifierList.php
nuest: `uniqueArray` is the internal object, sorting only happens on access.

api/v1/CodecheckApiHandler.php
nuest: Need to distinguish between write and read access > @dxL1nus creates an issue.

api/v1/CodecheckApiHandler.php
dxL1nus: I now renamed it to `CodecheckGithubRegisterApiClient` and the `JsonApiCaller` to `CodecheckApiClient`, as that is a lot more precise. Also the `CodecheckApiClient` is now extending my new `CurlApiClient`, that offers proper curl exception handling.

api/v1/CodecheckApiHandler.php
dxL1nus: the issue is created here: #87

dxL1nus:
> To avoid challenges with PRs that are based on each other, and still give me an opportunity to give feedback or review, you can also create a PR within your own fork:
> 
> https://github.com/dxL1nus/ojs-codecheck/compare/dxL1nus:feature/ui-retrieve-reserve-identifiers...dxL1nus:feature/unit-tests?expand=1
> 
> This makes the edits "on top of the other PR" readily accessible. I think I should be able to comment/review when you make me a collaborator on your repo.

I see. Should I do this in parallel to the ongoing PRs to this repo, so there is a second PR just on my fork which shows the differences between branches?

nuest (approved):

PR #83, opened 2026-01-04
feat: Add CODECHECK badge to issue table of contents

### LINKED ISSUE #15: Show CODECHECK on the issue's table of contents
- use small CODE WORKS badge
- identify the right hook to insert the information

subhanLabs:
Display badge for articles with completed CODECHECK using Templates::Issue::Issue::Article hook and Smarty template.

Closes #15
---

## Related User Stories

- A-CD1 (related): As an author, I want the CODECHECK certificate and badge to be displayed on my article's landing page...
- R-BV1 (related): As a reader, I want to see a CODECHECK badge on the article landing page and the issue's table of contents...


classes/migration/CodecheckSchemaMigration.php
```diff
@@ -10,34 +10,35 @@ class CodecheckSchemaMigration extends Migration
     public function up(): void
     {
         try {
-            if (Schema::hasTable('codecheck_metadata')) {
-                error_log("CODECHECK: Table exists, cleaning up old structure");
-                Schema::dropIfExists('codecheck_metadata');
-                error_log("CODECHECK: Dropped old table");
+            // Only create table if it doesn't exist
+            if (!Schema::hasTable('codecheck_metadata')) {
+                error_log("CODECHECK: Creating codecheck_metadata table");
```
nuest: Is there an alternative log function without a notion of "error" ?

classes/migration/CodecheckSchemaMigration.php
```diff
@@ -10,34 +10,35 @@ class CodecheckSchemaMigration extends Migration
...
+                
+                Schema::create('codecheck_metadata', function (Blueprint $table) {
+                    $table->bigInteger('submission_id')->primary();
+                    $table->string('version', 50)->default('latest');
```
nuest: I know this is unrelated to this PR, so independent from the review/fixes, but please clarify: is this the version of the specification that is used? I think the field name should be more informative: `spec_version`

If this is indeed the specification version, then please open a new issue to capture the task to rename the field.

classes/migration/CodecheckSchemaMigration.php
```diff
@@ -10,34 +10,35 @@ class CodecheckSchemaMigration extends Migration
...
+                    $table->text('summary')->nullable();
+                    $table->string('report', 500)->nullable();
+                    $table->text('additional_content')->nullable();
+                    $table->timestamps();
+                    $table->index('submission_id');
+                });
+                
+                error_log("CODECHECK: Table created successfully");
+            } else {
+                error_log("CODECHECK: Table already exists, skipping creation");
```
nuest: What about the migrations, for example if a field has a different name or a field was added? Are there mechanisms to handle that, or do we need to do this ourselves here? At the moment we don't need migrations, we can just rename, just wondering for the future if we need to do something more complex in this else clause.

classes/Submission/CodecheckSubmissionDAO.php
```diff
@@ -84,79 +67,166 @@ public function __construct(array $data)
...
+    {
+        $certificate = $this->getCertificate();
+        
+        // If it's already a URL, return it
+        if (filter_var($certificate, FILTER_VALIDATE_URL)) {
+            return $certificate;
         }
+        
+        // If it's a CODECHECK ID, build the URL
+        if (preg_match('/^CODECHECK-\d{4}-\d+$/', $certificate)) {
```
nuest, line 192: When are IDs saved with the preprended string `CODECHECK-` ? Is this an internal mechanism?

I would prefer to only store the certificate identifier as `YYYY-NNN`.

classes/Submission/CodecheckSubmissionDAO.php
```diff
@@ -84,79 +67,166 @@ public function __construct(array $data)
...
+    public function getDoiLink(): string 
+    {
+        $report = $this->getReport();
+        
+        // Check if report is a DOI
+        if (preg_match('/^(https?:\/\/)?(doi\.org\/)?(.+)$/', $report, $matches)) {
+            return 'https://doi.org/' . $matches[3];
         }
+        
         return '';
```
nuest: Can we add a detailed error log for this case? Maybe even alert the user?

We normally want a DOI for all checks, so when this code is reached, something is incomplete in the metadata.

_Why not return the content of `$report` as a fallback?_ Then the user can see the bad metadata and fix it.

classes/Submission/CodecheckSubmissionDAO.php
```diff
@@ -84,79 +67,166 @@ public function __construct(array $data)
...
         
-        // Fallback: extract last part of URL
-        $parts = explode('/', rtrim($url, '/'));
-        $lastPart = end($parts);
-        
-        if (strpos($lastPart, 'certificate-') === 0) {
-            return 'CODECHECK ' . str_replace('certificate-', '', $lastPart);
-        }
-        
         return 'View Certificate';
```
nuest: This string needs to be localised. Maybe you can instead define the link text in the template and the language files?

nuest (changes requested):
One related question (please turn into issues), one actual question, and one actionable content. Please ping me when updated!

classes/Submission/CodecheckSubmissionDAO.php
dxL1nus: Hey @nuest it is fine if we leave it like this for now, as my certificate identifier branch is still not pushed yet. So once it is, we can adapt Subhans code to it

classes/migration/CodecheckSchemaMigration.php
subhanLabs: Yes, you're correct. The version field stores the CODECHECK configuration specification version (e.g., 1.0, 2.0, or latest).
I agree that spec_version would be more descriptive. I'll have opened the issue #93.

classes/migration/CodecheckSchemaMigration.php
subhanLabs: In this current version, we are only creating the table if it does not already exist. Previously, the code dropped and recreated the tables on enable, which ensured that updated fields defined in the code were applied. However, this approach caused all existing data to be lost.

```
            if (Schema::hasTable('codecheck_metadata')) {
                error_log("CODECHECK: Table exists, cleaning up old structure");
                Schema::dropIfExists('codecheck_metadata');
                error_log("CODECHECK: Dropped old table");
            }
            
            Schema::create('codecheck_metadata', function (Blueprint $table) {
                $table->bigInteger('submission_id')->primary();
                $table->string('version', 50)->default('latest');
                $table->string('publication_type', 50)->default('doi');
                $table->text('manifest')->nullable();
                $table->string('repository', 500)->nullable();
                $table->text('source')->nullable();
                $table->text('codecheckers')->nullable();
                $table->string('certificate', 100)->nullable();
                $table->timestamp('check_time')->nullable();
                $table->text('summary')->nullable();
                $table->string('report', 500)->nullable();
                $table->text('additional_content')->nullable();
                $table->timestamps();
                $table->index('submission_id');
            });
```

I have an idea to address this: we can define an initial schema with the column names. In future plugin versions, if we add new fields or rename existing ones, we can use these new functions to update the schema accordingly:

```
    if (Schema::hasColumn('codecheck_metadata', 'version') && 
        !Schema::hasColumn('codecheck_metadata', 'spec_version')) {
        Schema::table('codecheck_metadata', function (Blueprint $table) {
            $table->renameColumn('version', 'spec_version');
        });
    }
    
    Schema::table('codecheck_metadata', function (Blueprint $table) {
        if (!Schema::hasColumn('codecheck_metadata', 'new_field_1')) {
            $table->string('new_field_1', 100)->nullable();
        }
        if (!Schema::hasColumn('codecheck_metadata', 'new_field_2')) {
            $table->text('new_field_2')->nullable();
        }
    });
```

This way, if the user already has tables, their data will not be deleted, because the initial condition will be false when the table already exists. In that case, the table will not be recreated; instead, existing columns can be renamed or new ones can be added if needed.

classes/Submission/CodecheckSubmissionDAO.php
subhanLabs: Done! Now returning the raw report value as fallback so users can see and fix invalid DOI formats.

```
    /**
     * Get DOI link if available
     */
    public function getDoiLink(): string 
    {
        $report = $this->getReport();
        
        // If report is empty, return empty
        if (empty($report)) {
            return '';
        }
        
        // Check if report is a valid DOI format
        if (preg_match('/^(https?:\/\/)?(doi\.org\/)?(.+)$/', $report, $matches)) {
            $doi = $matches[3];
            
            // Validate DOI format (should contain at least one slash, e.g., 10.xxxx/yyyy)
            if (strpos($doi, '/') !== false && preg_match('/^10\.\d+\//', $doi)) {
                return 'https://doi.org/' . $doi;
            }
        }
        
        // Return raw value as fallback
        return $report;
    }
```

classes/Submission/CodecheckSubmissionDAO.php
subhanLabs: Good catch! I have removed this function entirely and will directly handle in the template file:
```
            {if $certificateLink}
                <div class="value">
                    <a href="{$certificateLink|escape}" 
                    target="_blank"
                    title="{translate key='plugins.generic.codecheck.certificate.landingPage.title'}">
                        {translate key='plugins.generic.codecheck.certificate.prefix'} {$codecheckData->getCertificate()|escape}
                    </a>
                </div>
            {/if}
```

classes/migration/CodecheckSchemaMigration.php
subhanLabs: I have removed all error_log functions, as other plugins like [PID Manager](https://github.com/GaziYucel/pidManager/tree/stable-3_5_0) and [geoMetadata](https://github.com/TIBHannover/geoMetadata) are also not logging.

classes/migration/CodecheckSchemaMigration.php
nuest: I think this defensive strategy would be very good.

Did you see anything similar in existing plugins? This kind of data migrations?

Maybe we can base this on the plugin version?

nuest (approved):

PR #92, opened 2026-01-21
Documentation/build release

### LINKED ISSUE #88: Add npm created files to .gitignore, add instructions to create build/ realese in README

dxL1nus:
Closes #88;

- This PR adds a/ extends the simple section on how to correctly build a release for the our plugin.
- I have also added the build files to the `.gitignore` file, so we don't get any merge conflicts because of this in the future.
- I have also stated in the guide on the creation of a build release, that users should double check, if the build folder actually exists, because it is normally added to the `.gitignore` file and thus not pushed.

Feel vriew to view the changes to the README in this [demo](https://github.com/dxL1nus/ojs-codecheck/tree/documentation/build-release?tab=readme-ov-file#frontend-structure) :D


nuest (changes requested):
Looks all good to me, but it's worth testing at this point (because we're targeting a live demo) if the "Build and Package your Plugin" section from https://docs.pkp.sfu.ca/dev/plugin-guide/en/release also works, and what is created with the `pkp-plugin release ...` command. Please try that out.

Also, the release instructions should have steps for setting a version at the correct spot (not just "create release tag) and then how to upload the created package to the release on GitHub (not extensive instructions but a link to GitHub docs how to create a release with a locally build file).

dxL1nus:
> Looks all good to me, but it's worth testing at this point (because we're targeting a live demo) if the "Build and Package your Plugin" section from https://docs.pkp.sfu.ca/dev/plugin-guide/en/release also works, and what is created with the `pkp-plugin release ...` command. Please try that out.

I'll try that out and let you know what happens

dxL1nus:
> Also, the release instructions should have steps for setting a version at the correct spot (not just "create release tag)

Okay should I reference the [CHANGELOG.md](https://github.com/codecheckers/ojs-codecheck/blob/main/CHANGELOG.md) there?

> and then how to upload the created package to the release on GitHub (not extensive instructions but a link to GitHub docs how to create a release with a locally build file).

Makes sense, I will add this.

dxL1nus:
> Looks all good to me, but it's worth testing at this point (because we're targeting a live demo) if the "Build and Package your Plugin" section from https://docs.pkp.sfu.ca/dev/plugin-guide/en/release also works, and what is created with the `pkp-plugin release ...` command. Please try that out.

Looks like this doesn't work, because first of all, it doesn't run `npm build run` and second of all, it doesn't work with the required `version.xml` [schema](https://docs.pkp.sfu.ca/dev/plugin-guide/en/getting-started#versionxml)

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE version SYSTEM "../../../lib/pkp/dtd/pluginVersion.dtd">
<version>
    <application>tutorialExample</application>
    <type>plugins.generic</type>
    <release>1.0.0.0</release>
    <date>2023-05-15</date>
</version>
```

It doesn't work, because the plugin expects a tag `<version>` with the version directly in it so e.g.:

```xml
<version>1.0.0.0</version> <!-- What the pkp-plugin-cli expects -->
```

So it only builds a release, when that is inside the `version.xml`, which is against the conventions by OJS themselves. So I don't know what to do now. Do we want to just make a guide to manually create a release, or do we want to modify the `version.xml` to something it shouln't be as per the conventions and then be able to use the tool?

What is your opinion @nuest?

dxL1nus:
I think [this](https://github.com/dxL1nus/ojs-codecheck/blob/documentation/build-release/README.md#creating-a-release) is already looking better now. What could be improved further @nuest?

nuest:
> Okay should I reference the [CHANGELOG.md](https://github.com/codecheckers/ojs-codecheck/blob/main/CHANGELOG.md?rgh-link-date=2026-01-22T09%3A12%3A34Z) there?

Yes, good idea to not forget updating the CHANGELOG.

While you're at it, maybe do a quick search of release checklists of other OJS plugins? Just 5 Minutes, if something comes up, then great, otherwise we'll manage without that help.

nuest:
> I think [this](https://github.com/dxL1nus/ojs-codecheck/blob/documentation/build-release/README.md?rgh-link-date=2026-01-22T10%3A07%3A49Z#creating-a-release) is already looking better now. What could be improved further @nuest?

Yes, very useful.

1. Re "Package plugin:" can you provide a CLI command that does just that? Probably using the `zip` tool..
2. "git tag" command is missing the code formatting
3. "please also add a thorough description, on what this release contains" > suggest to use the entries from the changelog here as the basis
4. For release titles, let's come up with something fun!
  * My first idea: we use the last names of authors of papers that received a certificate, maximum 5, in order of the certificate numbers. The first release would therefore be named: "Lee, Suh, Hill & Piccolo".
  * What is your idea?

dxL1nus:
> While you're at it, maybe do a quick search of release checklists of other OJS plugins? Just 5 Minutes, if something comes up, then great, otherwise we'll manage without that help.

It seems that even plugins created by the PKP devs themselves (example [here](https://github.com/pkp/customBlockManager?tab=readme-ov-file)) don't have a CHANGELOG. That being said I think it is still good to keep it, or at least the part which is describing how we choose our version names, as this would be a good guide to other/ new developers.

dxL1nus:
> While you're at it, maybe do a quick search of release checklists of other OJS plugins? Just 5 Minutes, if something comes up, then great, otherwise we'll manage without that help.

Sorry I now read it again and see that your question was something different. So it also seems, that no plugin that I found so far has a guide on how to build a release at all in their README

dxL1nus:
>     1. Re "Package plugin:" can you provide a CLI command that does just that? Probably using the `zip` tool..

So I just found out, that GitHub automatically creates a `.zip` and a `.tar.gz` archive to download from a release when creating it. So I think the devs need no extra step in packaging the directory themselves here. Look at this [example release](https://github.com/dxL1nus/ojs-codecheck/releases/tag/v1.0.0.0) on my fork.

dxL1nus:
> > ```
> > 1. Re "Package plugin:" can you provide a CLI command that does just that? Probably using the `zip` tool..
> > ```
> 
> So I just found out, that GitHub automatically creates a `.zip` and a `.tar.gz` archive to download from a release when creating it. So I think the devs need no extra step in packaging the directory themselves here. Look at this [example release](https://github.com/dxL1nus/ojs-codecheck/releases/tag/v1.0.0.0) on my fork.

So I looked into it further and the only real reason why someone would want to package the repo themselves is to delete the `/resources` folder, but other than that it is basically the same. So I am wondering if could also just keep that folder in the release and use the automatically zipped folder of github instead, because this way we save the dev of one step in creating the release. What do you say? But I am of course also open to dicuss this in a meeting for example

dxL1nus:
>     4. For release titles, let's come up with something fun!
> 
> 
>     * My first idea: we use the last names of authors of papers that received a certificate, maximum 5, in order of the certificate numbers. The first release would therefore be named: "Lee, Suh, Hill & Piccolo".
> 
>     * What is your idea?

Hmm, I don't know if that would get confusing. Until now I'd just say for normal releases a simple `"Release x.y.z.0"` would be enough. And for special/ major releases something like `"Major Release 2.0.0.0"` or something like this.

Maybe we also classify it as `stable`, `beta`, `alpha`, etc. release.

nuest:
I propose to use both release number and a speaking title, side by side.

We can use terms like "beta" and "alpha" in the release description, and even in the release name, as classification, yes. "stable" not so much - everything is only stable until it becomes unstable.

nuest:
A reason to create your own zip file is that some of the files needed for a release are created during the build process, and these are not part of the codebase. The zip automatically created by GitHub includes all the source files, but not all files created by `npm run build`, which are needed when installing the plugin. I would like users (not developers) to be able to download an archive of the plugin, unzip it in their OJS, and then use the plugin.

We might also at some point provide different versions of the plugin for different OJS versions based on branches in the codebase - this would also require manual uploading of the release files to the GitHub release.

Hope this clarifies.

nuest (approved):

dxL1nus:
> I propose to use both release number and a speaking title, side by side.
> 
> We can use terms like "beta" and "alpha" in the release description, and even in the release name, as classification, yes. "stable" not so much - everything is only stable until it becomes unstable.

okay, I just thought we use `stable`, as this is what all the other PKP devs use for their Plugin releases and even the OJS releases itsself, but I get your point

dxL1nus:
> A reason to create your own zip file is that some of the files needed for a release are created during the build process, and these are not part of the codebase. The zip automatically created by GitHub includes all the source files, but not all files created by `npm run build`, which are needed when installing the plugin. I would like users (not developers) to be able to download an archive of the plugin, unzip it in their OJS, and then use the plugin.
> 
> We might also at some point provide different versions of the plugin for different OJS versions based on branches in the codebase - this would also require manual uploading of the release files to the GitHub release.
> 
> Hope this clarifies.

well yes that totally makes sense, I somehow forgot, that the built files aren't pushed to GitHub, so this isn't included in the automatically created archive.

dxL1nus:
> > A reason to create your own zip file is that some of the files needed for a release are created during the build process, and these are not part of the codebase. The zip automatically created by GitHub includes all the source files, but not all files created by `npm run build`, which are needed when installing the plugin. I would like users (not developers) to be able to download an archive of the plugin, unzip it in their OJS, and then use the plugin.
> > We might also at some point provide different versions of the plugin for different OJS versions based on branches in the codebase - this would also require manual uploading of the release files to the GitHub release.
> > Hope this clarifies.
> 
> well yes that totally makes sense, I somehow forgot, that the built files aren't pushed to GitHub, so this isn't included in the automatically created archive.

Now I noticed, that the archive commits don't work for this, as they don't include the public folder. So I have to redo that section of the documentation. I am sorry that I only notice that now.

dxL1nus:
> Now I noticed, that the archive commits don't work for this, as they don't include the public folder. So I have to redo that section of the documentation. I am sorry that I only notice that now.

I now updated this [here](https://github.com/dxL1nus/ojs-codecheck/blob/documentation/build-release/README.md#creating-a-release). It should now include all relevant build files.

nuest:
Well, it's hard to get things right the first time.

Will you open a new PR, please?

I also have a follow up question regarding step "Prepare the repository for packaging: " - Do you expect the developer to delete these folders? Could we not manage this exclusion through parameters of the `git archive` and `zip` commands?

dxL1nus:
> I also have a follow up question regarding step "Prepare the repository for packaging: " - Do you expect the developer to delete these folders? Could we not manage this exclusion through parameters of the `git archive` and `zip` commands?

yes this is now already done with the commands. It was still included in this guide, from Subhans initial version, but you are correct, that we don't need this anymore.

dxL1nus:
> Well, it's hard to get things right the first time.
> 
> Will you open a new PR, please?

I created the new pull request #95

PR #97, opened 2026-02-25
feat: restore wizard manager and review refresher for custom field persistence

subhanLabs:
This feature was present in previous commits but was removed and is required to ensure custom submission fields are saved in the database.

- Add CodecheckWizardManager to load/save custom submission fields (codeRepository, dataRepository, manifestFiles, dataAvailabilityStatement) via API on wizard Continue/Save for Later button clicks
- Add CodecheckReviewRefresher to detect review step and refresh CODECHECK panel with live data from the API
- Ensure Vue components mount after saved data is loaded into textareas so they receive correct initial values
---

## Related User Stories

- A-ME1 (related): As an author, I want to enter CODECHECK-related metadata (code URL, data URL, expected outputs) only once...


resources/js/main.js
```diff
@@ -153,10 +153,281 @@ pkp.registry.storeExtend("fileManager_SUBMISSION_FILES", (piniaContext) => {
...
+
+  getSubmissionId() {
+    const match = window.location.search.match(/id=(\d+)/);
+    return match ? match[1] : null;
+  }
+
+  setupButtonListener() {
+    document.addEventListener('click', (e) => {
+      const button = e.target.closest('button');
+      if (button && (button.textContent.includes('Continue') || button.textContent.includes('Save for Later'))) {
```
nuest: Will this work if another language is used?

resources/js/main.js
```diff
@@ -153,10 +153,281 @@ pkp.registry.storeExtend("fileManager_SUBMISSION_FILES", (piniaContext) => {
...
+    });
+  }
+
+  isOnReviewStep() {
+    const wizardSteps = document.querySelectorAll('.submissionWizard__step');
+    
+    for (const step of wizardSteps) {
+      if (step.classList.contains('isActive') || step.classList.contains('isCurrent')) {
+        const stepText = step.textContent.toLowerCase();
+        if (stepText.includes('review') || stepText.includes('submit')) {
```
nuest: Same here: will this work in other languages?

resources/js/main.js
```diff
@@ -153,10 +153,281 @@ pkp.registry.storeExtend("fileManager_SUBMISSION_FILES", (piniaContext) => {
...
+
+      body.innerHTML = '';
+      
+      let hasData = false;
+      
+      if (publication.codeRepository) {
+        hasData = true;
+        body.innerHTML += `
+          <div class="submissionWizard__reviewPanel__item">
+            <h4>${this.escapeHtml('Code Repository URLs')}</h4>
```
nuest: This string needs to be localized. Please check other occurences of `.escapeHtml(..)`, too.

resources/js/main.js
```diff
@@ -153,10 +153,281 @@ pkp.registry.storeExtend("fileManager_SUBMISSION_FILES", (piniaContext) => {
...
+            </div>
+          </div>
+        `;
+      }
+      
+      if (publication.dataRepository) {
+        hasData = true;
+        body.innerHTML += `
+          <div class="submissionWizard__reviewPanel__item">
+            <h4>${this.escapeHtml('Data Repository URLs')}</h4>
```
nuest: Localize string.

nuest (changes requested):

resources/js/main.js
subhanLabs: You were right, I have change it to use the button ID, which I tested remain same in other languages

```
  setupButtonListener() {
    document.addEventListener('click', (e) => {
      const button = e.target.closest('button');
      if (!button) return;
      
      // Save on any button click except cancel
      if (button.id !== 'cancelSubmission') {
        this.saveData();
      }
    }, true);
  }
```

resources/js/main.js
subhanLabs: I have replaced with the class of review if his parent class contains an active state.

```
  isOnReviewStep() {
    // Find the review step container
    const allSteps = document.querySelectorAll('.pkpStep');
    
    for (const step of allSteps) {
      // Check if this step contains review panels AND is not hidden
      const hasReviewPanels = step.querySelectorAll('.submissionWizard__reviewPanel').length >= 3;
      const isVisible = !step.hasAttribute('hidden');
      
      if (hasReviewPanels && isVisible) {
        return true;
      }
    }
    
    return false;
  }
```

nuest (approved):

PR #98, opened 2026-02-27
Add test dataset

### LINKED ISSUE #45: Add demo data #33
Based on articles from the CODECHECK register: https://codecheck.org.uk/register/

- [ ] one article with two codecheckers
- [ ] three articles from three different codecheckers
- [ ] three articles from three different conferences/venues
- [ ] put them in one journal, but two issues
- [ ] 5 demo articles are published
- [ ] 2 demo articles are under review
- [ ] 1 demo article is submitted

(there is some overlap between these criteria)

Use the same approach as here, but only use _one_ PDF for the submissions (duplicate as part of a script, or duplicate locally and ignore copies): https://github.com/TIBHannover/geoMetadata/tree/main/testData

subhanLabs:
Closes #45 

Adds OJS 3.5 test dataset following PKP datasets repository pattern.

Includes:
- 8 articles from CODECHECK register
- 5 published, 2 under review, 1 submitted
- Demonstrates multiple codecheckers and conference venues
- Database dump, files, and config ready for PKP datasets integration


testData/stable-3_5_0-codecheck/mysql/database.sql
nuest, line 1: This file is WAY to big. 26 MB!

It seems like it inlcudes the whole ROR database. That is not necessary and should be the first approach to reduce the size considerably.

nuest (changes requested):
One critical change only: reduce the size of the SQL file _a lot_.

This PR would also be nicely illustrated with a screenshot of the issue and article landing pages showing the test data. Please!

subhanLabs:
Explanation of what happened with this test dataset PR and why a new one has been opened.

After your feedback about the SQL file being 26MB, I reduced the size by removing the unnecessary ROR database, event_log, and email_log data. The file is now 400KB.

However, I also wanted to clean the large file from the git history entirely so it would not remain in previous commits. To do this I used `git filter-repo`, which removes a file from all previous commits. What I did not realize at the time is that this tool rewrites the commit hashes of every single commit in the repository. When I force pushed the branch to GitHub, the new commit hashes no longer matched the history of main, which caused GitHub to automatically close the pull request. I was not able to reopen it because the branch history was no longer compatible.

To keep things clean, I created a fresh branch from main with the same name and the 26MB file never appears in the new history at all.

I have opened a new pull request. I understand the previous review comments are lost. I have already addressed your main concern about the SQL file size and will also be adding screenshots of the OJS issue and article landing pages showing the test data as you requested.

Sorry for the inconvenience.

nuest:
@Subhanaliweb that is a good resolution of the problem. Thanks for explaining and documenting the details!

PR #107, opened 2026-03-24
Feature/yaml validation

### LINKED ISSUE #66: Add YAML-validation
After saving the CODECHECK metadata form (cf. #59) the YAML file should be internally created once and it should be validated, communicating errors back to the user.

This is especially useful when we allow arbitrary content, see https://codecheck.org.uk/spec/config/1.0/#additional-content

It could be that just parsing the file once with https://www.php.net/manual/en/function.yaml-parse.php will already work as a validation.

Or https://symfony.com/doc/current/components/yaml.html#clear-error-messages ?

dxL1nus:
Closes #66;

This PR is for the structured YAML Validation (syntax only for now, no semantic validation yet), when the CODECHECK Metadata Form is saved. On save, the created `codecheck.yml` file content gets sent to the PHP script which validates it and sends true or false, depending if it is valid YAML Syntax.

- after saving the CODECHECK metadata form the YAML file is internally created once and its contents are retrieved
- it then gets structurally validated (parsing the file once with [Symfony YAML](https://symfony.com/doc/current/components/yaml.html)), communicating errors back to the user
- the user only gets to download the file if there are no errors
- fully unit-tested as well

- - -

When the yaml file has wrong syntax, the concrete error message gets displayed to the user:

Image

This is for the example of the following invalid YAML file:

```yaml
this:
   >is:
       invalid: |
            yaml
            content.
```

- - -

The Unit Tests for the [YAML Structure Validator](https://github.com/dxL1nus/ojs-codecheck/blob/feature/yaml-validation/classes/Workflow/CodecheckYamlValidator.php) have 100% coverage:

Image

Image
---

## Related User Stories

- C-AD1 (related): As a codechecker, I want the plugin to automatically reserve a certificate identifier, create the Register issue...
- Y-PV1 (related): As a copyeditor, I want to validate the CODECHECK metadata for completeness...


dxL1nus:
@nuest could you please review this PR so it can be merged as soon as possible?

nuest:
I would prefer to review after the merge conflicts are resolved - is there a reason to review beforehand?

dxL1nus:
> I would prefer to review after the merge conflicts are resolved - is there a reason to review beforehand?

no there is no reason. I will resolve the merge conflicts

dxL1nus:
> I would prefer to review after the merge conflicts are resolved - is there a reason to review beforehand?

The branch is now up to date and can be reviewed for merging

classes/Workflow/CodecheckYamlValidator.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->isValidYaml = false;
+        $this->yamlParseException = new ParseException("The Yaml File wasn't validated yet.");
+    }
+
+    public function validateYaml(): void
+    {
+        try {
+            Yaml::parse($this->yamlContent);
+            $this->isValidYaml = true;
+            $this->yamlParseException = new ParseException("");
```
nuest: I don't really like this, the value of the exception should not be set if there was no exception. Please use a `null` value here instead, already at the construction stage.

The text of an Exception should not be used to communicate state.

locale/en/locale.po
```diff
@@ -239,6 +242,12 @@ msgstr "Are you sure you want to remove this identifier?"
 msgid "plugins.generic.codecheck.previewYaml"
 msgstr "Preview CODECHECK metadata file"
 
+msgid "plugins.generic.codecheck.yaml.valid"
+msgstr "CODECHECK metadata creates a structurally valid codecheck.yml file"
+
+msgid "plugins.generic.codecheck.yaml.invalid"
+msgstr "CODECHECK metadata creates a structurally invalid codecheck.yml file."
```
nuest: These two message are two close to each other, it can be hard to miss the `in`.

Suggestion for valid, shorter "codecheck.yml is valid."

For the invalid case, can you point out what user can do? When/where is this message shown? Maybe adding "Please check the detailed error message below." makes sense?

resources/js/Components/CodecheckMetadataForm.vue
```diff
@@ -832,6 +832,23 @@ export default {
         const data = await response.json();
         const yamlContent = data.yaml;
         
+        return yamlContent;
+        
+      } catch (error) {
```
nuest, line 837: Only catch when you are actually doing something with the error, for example logging. Otherwise remove and let the outer function handle it (like it has to do now anyway).

resources/js/Components/CodecheckMetadataForm.vue
```diff
@@ -832,6 +832,23 @@ export default {
...
+    },
+
+    async previewYaml() {
+      try {
+        const yamlContent = await this.generateYamlContent();
+        
+        let isValidYaml = await this.validateGeneratedYamlFile(yamlContent);
+
+        if (!isValidYaml) {
+          return;
```
nuest, line 850: How is the error then communicated to the user?

resources/js/Components/CodecheckMetadataForm.vue
nuest: Some information from the caught error should be included in the message here.

tests/WorkflowUnitTests/CodecheckYamlValidatorUnitTest.php
```diff
@@ -0,0 +1,67 @@
...
+    protected function setUp(): void
+	{
+		parent::setUp();
+
+        $this->codecheckYamlValidator = new CodecheckYamlValidator("example: {yaml: true}");
+	}
+
+    public function testYamlValidatorContstructor()
+    {
+        $expectedExceptionMessage = "The Yaml File wasn't validated yet.";
```
nuest: The error should just be null here.

CodecheckPlugin.php
nuest, line 1: Is this file really related to the yaml validation?

nuest (changes requested):

classes/Workflow/CodecheckYamlValidator.php
dxL1nus: I entirely rewrote it. It is much simpler now. If the YAML parse fails, the exception  is thrown and then forwarded, so the API Handler can catch the exception. Otherwise there is just no exception thrown, so the API Handler will return with a successful validation.

```php
public function validateYaml(): void
{
    try {
        Yaml::parse($this->yamlContent);
    } catch (ParseException $e) {
        throw $e;
    }
}
```

```php
public function validateYamlStructure(): void
{
    $postParams = json_decode(file_get_contents('php://input'), true);
    $yamlContent = $postParams["yaml"];

    $yamlValidator = new CodecheckYamlValidator($yamlContent);

    try {
        $yamlValidator->validateYaml();
    } catch (\Throwable $e) {
        error_log("[CODECHECK Api Handler] YAML Parse Exception: " . $e->getMessage());

        JsonResponse::staticResponse([
            'success' => false,
            'error' => $e->getMessage(),
        ], $e->getCode());
    }

    error_log("[CODECHECK Api Handler] The generated YAML content is structurally valid!");

    JsonResponse::staticResponse([
        'success' => true,
    ], 200);
}
```

locale/en/locale.po
dxL1nus: Now the error message looks like this:

grafik

It is shown below the form, where all error messages are shown, whenever the user wants to preview the YAML file, but its structure is invalid.

grafik

resources/js/Components/CodecheckMetadataForm.vue
dxL1nus: this is Subhans code. I was only assigned the yaml validation. This function is for the actual YAML creation / generation. I added logging here, as I think in principle it is good to keep the catch statement here.

resources/js/Components/CodecheckMetadataForm.vue
dxL1nus: from inside the function `this.validateGeneratedYamlFile(yamlContent)` directly. This function is just called here.

```php
async validateGeneratedYamlFile(yamlContent) {
  try {
    console.log('Validating the created codecheck.yml file');
    
    let apiUrl = pkp.context.apiBaseUrl;

    ...

    if (data.success) {
        console.log(this.t('plugins.generic.codecheck.yaml.valid'), 'success');
        return true;
    } else {
        console.error('Structural Validation error:', data.error);
        this.showMessage(`${this.t('plugins.generic.codecheck.yaml.invalid')}\n${this.t('plugins.generic.codecheck.error')}: ${data.error}`, 'error');
        return false;
    }
  } catch (error) {
    console.error('Structural Validation API fetch error:', error);
    this.showMessage(this.t('plugins.generic.codecheck.yaml.invalid'), 'error');
    return false;
  }
},
```

resources/js/Components/CodecheckMetadataForm.vue
dxL1nus: I did it like this: 
```this.showMessage(`${this.t('plugins.generic.codecheck.yamlPreviewFailed')}\n${error}`, 'error');```

File: tests/WorkflowUnitTests/CodecheckYamlValidatorUnitTest.php (hunk 5 above)
Comment (dxL1nus) [reply tests/WorkflowUnitTests/CodecheckYamlValidatorUnitTest.php]: this entire test has just become obsolete anyways, because of the refactor from the first conversation, so I removed it entirely

File: classes/Workflow/CodecheckYamlValidator.php (hunk 1 above)
Comment (dxL1nus) [reply classes/Workflow/CodecheckYamlValidator.php]: I've simplified it even more to just this:

```php
public function validateYaml(): void
{
    // This will throw a Symphony ParseException, if the YAML content is invalid
    Yaml::parse($this->yamlContent);
}
```

CodecheckPlugin.php
dxL1nus: no this was just my try to find a hook to activate the yaml validation in the copyediting phase automatically. But since I haven't found any hook to do this so far and since we agreed on not doing this functionality in this PR anyways, I will remove this

dxL1nus:
I have resolved all conversations and rebased this branch to the latest version of main so it should now be ready to be merged! @nuest

nuest (approved):

PR #79, opened 2025-12-18
Feature/unit tests

dxL1nus:
closes #70
closes #40 
closes #73 
closes codecheckers/check-pub#35
closes #96

**!IMPORTANT:** This PR can only be safely merged after PR #63 

This is the PR for the first unit tests for the ojs-codecheck plugin. The tests are mainly for the RetrieveReserveIdentifiers scripts and can act as a guideline/ template on how to write other unit tests for the remaining class files.

Run the tests with the following command from the PKP OJS root directory:
```bash
lib/pkp/lib/vendor/phpunit/phpunit/phpunit -c lib/pkp/tests/phpunit.xml plugins/generic/codecheck/tests/
```

Results:
grafik

find the test coverage in the following file:
```bash
/var/www/html/ojs-tests/lib/pkp/tests/results/index.html
```

grafik

PR #111, opened 2026-04-07
Feature/issue 72 yaml generation test

### LINKED ISSUE #72: Add a UI test that ensures the YAML generated for the preview (JS) matches the YAML generated for download (PHP)
---

## Related User Stories

- Story 13 (related): As a codechecker, I want the plugin to provide a form that generates, validates, and stores the codecheck.yml...

---

## Related User Stories

subhanLabs:
Closes #72 

!IMPORTANT: This PR can only be safely merged after PR [#79](https://github.com/codecheckers/ojs-codecheck/pull/79)

## Overview
This PR establishes PHP as the single source of truth for YAML generation. Previously, the preview button used JavaScript (`js-yaml`) while the download button used PHP, which could lead to inconsistencies. Now both use the same PHP endpoint, ensuring identical output.

### Backend
- Updated `CodecheckMetadataHandler.php` to handle YAML generation with proper formatting
- Added Symfony YAML library
- Fixed date format to match original YAML specifications (`YYYY-MM-DD HH:MM:SS`)

### Frontend
- Removed `js-yaml` dependency from the Vue component
- Preview button now calls the PHP `/yaml` endpoint instead of generating YAML client-side
- Added "unsaved changes" tracking
- Added helpful tooltips explaining when preview is disabled

### Testing
- Added E2E test verifying preview/download consistency
- Test works dynamically with any published submission (no hardcoded data)
- Updated documentation with E2E testing instructions

## Testing
```bash
# Run E2E test locally
npm run test:e2e
```

The test passes ✅ - Preview YAML exactly matches Download YAML
image
---

## Related User Stories


cypress/tests/e2e/yaml-generation.cy.js
nuest, line 1: This is a good test for when "everything works".

It would be good to add two more tests that check the errors for "no preview because unsaved changes" and for "no preview because fields missing".

composer.json
```diff
@@ -3,11 +3,15 @@
...
-        "vlucas/phpdotenv": "^5.6"
+        "vlucas/phpdotenv": "^5.6",
+        "symfony/yaml": "^6.0"
     },
     "config": {
         "allow-plugins": {
             "php-http/discovery": true
+        },
+        "platform": {
+            "php": "8.2.26"
```
nuest: Can you make this "8.2 or higher"? It feels like a specific bugfix version could cause problems for other developers, but I also don't know what the platform here is used for. If you research results in "this is not critical", then leave it as is.

cypress.config.js
```diff
@@ -12,5 +12,10 @@ export default defineConfig({
     specPattern: 'cypress/tests/component/**/*.cy.js',
     supportFile: 'cypress/support/component.js',
     indexHtmlFile: 'cypress/support/component-index.html'
+  },
+    e2e: {
+    baseUrl: process.env.CYPRESS_BASE_URL || 'http://localhost:8888/ojs',
```
nuest, line 19: Where is `:8888/ojs` configured? Is that the default from CI? Please add a line comment here for clarification.

nuest (changes requested):
Good work - I'm only missing two test cases for when the preview does not work because of missing fields or because of unchanged changes.

cypress/tests/e2e/yaml-generation.cy.js
subhanLabs: Added! I've included two additional test cases:
1. Preview button disabled when required fields are missing (removes certificate and verifies button is disabled)
2. Preview button disabled when unsaved changes exist (edits a field and verifies button is disabled)

`Note:` These tests require CODECHECK metadata to exist on the submission. They will pass once test data is present.

composer.json
subhanLabs: Changed to `"php": "8.2"`. 

The platform config tells Composer which PHP version to assume when resolving dependencies, even if you're running a different version locally. Setting it to `"8.2"` ensures we install packages compatible with PHP 8.2+ (OJS 3.5's minimum requirement).

This prevents accidentally installing dependencies that require PHP 8.3+ features while still allowing the project to run on any 8.2.x version.

cypress.config.js
subhanLabs: Added comment explaining the baseUrl:
// Default: local OJS dev setup (port 8888)
// Override with: CYPRESS_BASE_URL=http://localhost:3000/ojs npm run test:e2e

nuest (approved):

PR #109, opened 2026-03-24
Feature/Settings Form

dxL1nus:
Closes #89,
closes #86,
closes #91,
closes #102,
closes #81,
closes #90,
closes #106,
closes #17;

This is the current state of the Plugin Settings Form. All Settings seen here, are functional / working, but it is not yet fully finished, hence why it is a draft.

grafik
grafik
---

## Related User Stories

- P-CS1 (related): As a journal manager, I want to configure the CODECHECK mode, submission form fields, and instructions...
- P-AS1 (related): As a publisher, I want to enable CODECHECK by installing the plugin and configuring settings...
- P-CS2 (related): As a publisher, I want to configure the CODECHECK badge image, logo, and placement...

_User story references from the [CHECK-PUB User Stories document](https://codecheck.org.uk/pub/)._

PR #126, opened 2026-06-01
feat: ORCID deposition for codecheckers + CODECHECK form in reviewer tab (#16)

### LINKED ISSUE #16: Codechecking activity is deposited in ORCID profile
**User story:**

- _As a codechecker, I want my contributions to be publicly visible in established academic profiles (e.g., ORCID, Google Scholar, Web of Science, Semantic Scholar), so that I get recognition for my efforts and expertise._
- _As a journal editor, I want to be able to ensure recognition for a CODECHECK, no matter if conducted within OJS or outside, so that I can provide the same benefits to all codecheckers._

------

This requires to become a registered entity with ORCID, probably. ~~Not sure if every journal has to do that, or if we need an intermediary.~~ A journal needs to register a "Member API Client".

The option must be enabled in the settings.

Check out the ORCID plugin - how does it talk to the ORCID API, and how does it handle the authentication with the API on behalf of the journal: https://docs.pkp.sfu.ca/orcid/en/installation-setup

Start all development with the ORCID dev/test instance.
https://info.orcid.org/documentation/workflows/peer-review-workflow/

It should be possible to (re)trigger this deposition manually. The usual case should be that this is happening when the checked article is published.

See the ORCID plugin for inspiration / help: https://docs.pkp.sfu.ca/orcid/en/
---

## Related User Stories

- **C-RA1** (direct): As a codechecker, I want my CODECHECK contributions to be automatically deposited in academic profiles...
- **C-RA2** (direct): As a codechecker, I want my CODECHECK contribution to be recorded as a peer-review activity on my ORCID page...
- **E-EI1** (direct): As an editor, I want to import an externally completed CODECHECK...

subhanLabs:
## Summary
This PR implements two major features for the CODECHECK plugin:

1. **ORCID Deposition** — allows codecheckers to authorise the journal to deposit their CODECHECK peer-review activity to their ORCID profile via the ORCID Member API.
2. **CODECHECK form in reviewer tab** — injects a collapsible CODECHECK documentation section into the reviewer's Download & Review tab (tab 3).

### New Classes
- `classes/Orcid/OrcidApiClient.php` — low-level ORCID API HTTP client
- `classes/Orcid/OrcidAuthHandler.php` — handles OAuth 2.0 authorisation flow (startAuth/callback)
- `classes/Orcid/OrcidDepositService.php` — builds and deposits peer-review records to ORCID
- `classes/Orcid/OrcidTokenDAO.php` — manages token storage in `codecheck_orcid_tokens` table
- `classes/Orcid/PeerReviewPayloadBuilder.php` — constructs ORCID peer-review XML payload
- `classes/Log/CodecheckLogger.php` — structured logger with debug/info/error levels

### Modified Files
- `CodecheckPlugin.php` — ORCID hooks, reviewer data injection, publish handler
- `api/v1/CodecheckApiHandler.php` — added `orcid-status` and `orcid-deposit` endpoints, reviewer role
- `api/v1/JsonResponse.php` — made constructor args optional, added `response()` convenience method
- `classes/Constants.php` — added all ORCID and GitHub constants
- `classes/migration/CodecheckSchemaMigration.php` — added `codecheck_orcid_tokens` table
- `locale/en/locale.po` — added ORCID and missing locale keys
- `registry/uiLocaleKeysBackend.json` — added ORCID locale keys
- `resources/js/main.js` — ORCID section in workflow, reviewer form injection via MutationObserver
- `resources/js/Components/CodecheckOrcidSection.vue` — Vue component for ORCID authorisation and deposit UI

### ORCID OAuth flow
1. Editor opens the CODECHECK tab in the editorial workflow
2. Each codechecker sees an "Authorise ORCID" button
3. Clicking redirects to ORCID sandbox/production for consent
4. On callback, token is stored in `codecheck_orcid_tokens`
5. On publication (or manually), deposit is triggered automatically

### Reviewer form
1. When reviewer opens tab 3 (Download & Review), a MutationObserver detects the tab content loading
2. A collapsible "CODECHECK Documentation" section is injected below the review form
3. It mounts `CodecheckMetadataForm` and optionally `CodecheckOrcidSection` if ORCID is enabled

## Notes
- ORCID settings (client ID, secret, API type) will appear in the settings page once the settings form [PR](https://github.com/codecheckers/ojs-codecheck/pull/109) is merged
- Sandbox API is used by default during development
- All raw `error_log()` calls replaced with `CodecheckLogger` (debug/info/error levels)

Closes #16


nuest:
Thanks!

Can you please clarify re. "Editor opens the CODECHECK tab in the editorial workflow" > the codechecker needs to be the one that opens the tab and initiates the authentication with ORCID, right?

It would be great, because this is a crucial feature, if you could add a couple of screenshots or maybe even a screencast to document the completed functionality.

classes/migration/CodecheckSchemaMigration.php
nuest, line 1: Does this mean we store the token _forever_ and can edit the profile with that token?

Is there a way for a codechecker to stop allowing us to add data to their profile?

classes/Orcid/OrcidApiClient.php
nuest, line 1: How does this file relate to the ORCID API client functionality in the ORCID plugin? Is is functionally equivalent, was there re-use of ideas?

classes/Orcid/OrcidApiClient.php
```diff
@@ -0,0 +1,312 @@
...
+        );
+    }
+
+    /**
+     * POST a peer-review item to the codechecker's ORCID record.
+     * Returns the put-code string on success.
+     *
+     * @throws \RuntimeException on failure
+     */
+    public function postPeerReview(string $orcidId, string $accessToken, array $payload): string
```
nuest, line 171: How is the successful posting stored in the OJS system? When a codechecker visits the page a second time, is there still a "add to profile" button?

classes/Orcid/PeerReviewPayloadBuilder.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        $payload = [
+            'reviewer-role'          => 'reviewer',
+            'review-type'            => 'review',
+            'review-completion-date' => $this->buildDate($checkDate),
+            'review-group-id'        => $groupId,
+            'review-identifiers'     => $this->buildReviewIdentifiers($certificateDoi),
+            'convening-organization' => $this->buildConveningOrganization($journal),
+            'subject-type'           => 'journal-article',
```
nuest, line 51: What subject types does the ORCID API know? Could there be the need here to switch to something else, like a preprint?

classes/Orcid/PeerReviewPayloadBuilder.php
```diff
@@ -0,0 +1,161 @@
...
+        $issn = !empty($journal['issn']) ? trim($journal['issn']) : '';
+        if (!empty($issn)) {
+            $groupId = 'issn:' . $issn;
+        } else {
+            $groupId = 'orcid-generated:codecheck-ojs';
+        }
+
+        $payload = [
+            'reviewer-role'          => 'reviewer',
+            'review-type'            => 'review',
```
nuest, line 46: What review types does the ORCID API know?

classes/Orcid/PeerReviewPayloadBuilder.php
```diff
@@ -0,0 +1,161 @@
...
+            ? trim($journal['publisherName'])
+            : (!empty($journal['name']) ? trim($journal['name']) : 'CODECHECK Journal');
+
+        $city    = !empty($journal['publisherCity'])    ? trim($journal['publisherCity'])    : 'Unknown';
+        $country = !empty($journal['publisherCountry']) ? trim($journal['publisherCountry']) : 'XX';
+
+        // Ensure none are empty strings (ORCID rejects them)
+        if (empty($publisherName)) $publisherName = 'CODECHECK Journal';
+        if (empty($city))          $city           = 'Unknown';
+        if (empty($country))       $country        = 'XX';
```
nuest: We do not want to publish that kind of "fallback information" in a database that will keep this forever.

What is your suggestion to solve this?

I'm happy, for example, if there is an error message and a deactivated "Deposit" button if city and country are not properly configured, with a warning message that points to the editors to fix this.

Maybe we can even have a little "test ORCID deposition" buttong on the settings pages that A) checks if all required journal-wide metadata is there, and B) makes an authenticated request (that does not write anything) to confirm the set-up is working.

classes/Constants.php
```diff
@@ -26,7 +26,37 @@ class Constants
...
+    public const ORCID_CLIENT_SECRET = 'orcidClientSecret';
+
+    // ORCID API type values
+    public const ORCID_API_TYPE_SANDBOX = 'memberSandbox';
+    public const ORCID_API_TYPE_PRODUCTION = 'member';
+
+    // ORCID API base URLs
+    public const ORCID_URL_SANDBOX = 'https://sandbox.orcid.org';
+    public const ORCID_URL_PRODUCTION = 'https://orcid.org';
+    public const ORCID_API_URL_SANDBOX = 'https://api.sandbox.orcid.org/v3.0';
```
nuest, line 102: how can an editor (e.g., for testing) switch between the API endpoints?

resources/js/Components/CodecheckOrcidSection.vue
```diff
@@ -0,0 +1,324 @@
...
+        v-for="cc in codecheckers"
+        :key="cc.orcidId || cc.name"
+        class="codecheck-orcid-row"
+      >
+        <!-- Identity -->
+        <div class="codecheck-orcid-row__identity">
+          <strong>{{ cc.name }}</strong>
+          <span v-if="cc.orcidId" class="codecheck-orcid-row__orcid-id">
+            <img
+              src="https://orcid.org/sites/default/files/images/orcid_16x16.png"
```
nuest: Let's ship this file with our plugin.

changes.diff
nuest, line 1: Should this file be actually added here?

CodecheckPlugin.php
```diff
@@ -60,32 +67,99 @@ public function register($category, $path, $mainContextId = null): bool
...
+            Hook::add('Publication::publish', $this->onPublicationPublish(...));
         }
 
         return $success;
     }
+
+    /**
+     * Triggered when an editor publishes an article.
+     */
+    public function onPublicationPublish(string $hookName, array $args): bool
```
nuest, line 192: This ssems to require that the ORCID credentials must be provided by codecheckers before the publication.

_Is it also possible that a CODECHECK goes to the OJS backend after the article is published and then triggers the deposition?

CodecheckPlugin.php
```diff
@@ -60,32 +67,99 @@ public function register($category, $path, $mainContextId = null): bool
...
+        if (!$submission->getData('codecheckOptIn')) return false;
+
+        $context = Application::get()->getRequest()->getContext();
+        if (!$this->getSetting($context->getId(), Constants::ORCID_ENABLED)) return false;
+
+        try {
+            $depositService = new OrcidDepositService($this);
+            $results = $depositService->depositForSubmission($submission->getId());
+            foreach ($results as $result) {
+                if ($result['status'] === 'success') {
```
nuest, line 207: Please make a proposal how to include user facing messaging about the results and failures here.

If that means we need a database table to capture the deposition state, then that is fine.

CodecheckPlugin.php
```diff
@@ -121,12 +195,36 @@ public function callbackTemplateManagerDisplay($hookName, $args): bool
     {
         $templateMgr = $args[0];
         $request = Application::get()->getRequest();
-        
-        if ($request->getRequestedOp() == 'workflow') {
+
+        // ----------------------------------------------------------------
+        // Editor workflow page
```
nuest: "Start auth" must be triggered by codechecker, not by editor.

nuest (dismissed):
General comments first, detailed line-by-line review later.

classes/migration/CodecheckSchemaMigration.php
subhanLabs: **Token storage:**
Yes we store the token, but the scope is limited to `/activities/update`, we can only add/update peer-review records, nothing else on their profile.

**Revoking access:**
The codechecker must authorise themselves, we match their ORCID ID against the authenticated token, so no one else can authorise on their behalf. To stop allowing deposits, they can revoke access directly from their ORCID account under "Trusted organizations" which immediately invalidates our stored token.

classes/Orcid/OrcidApiClient.php
subhanLabs: The OJS ORCID plugin is not installed by default, so we could not rely on it being available. To keep our plugin self-contained and working out-of-the-box for all journals without additional dependencies, we implemented a lightweight ORCID client covering only what we need, the OAuth 2.0 flow and peer-review deposition via the Member API. The general approach follows the same ORCID documentation patterns.

classes/Orcid/OrcidApiClient.php
subhanLabs: image

After a successful deposit, we store the ORCID `put-code` returned by the ORCID API along with `deposit_status = 'success'` in the `codecheck_orcid_tokens` table. 

When the codechecker visits the page again, the Vue component loads the current status via the `orcid status` API endpoint which reads from this table. If `deposit_status` is `success`, the button changes from "Deposit to ORCID" to "Re-deposit to ORCID" — so they can trigger it  again if needed (e.g. if the metadata changed) but it is clear the deposition already happened. The put-code is also used to show a "View on ORCID" link directly to their peer-review record.

classes/Orcid/PeerReviewPayloadBuilder.php
subhanLabs: The ORCID peer-review API supports these subject types: `book`, `book-chapter`, `book-review`, `dataset`, `edited-book`, `encyclopedia-entry`, `journal-article`, `journal-issue`, `magazine-article`, `manual`, `newsletter-article`, `newspaper-article`, `online-resource`, `other`, `preprint`, `report`, `research-tool`, `supervised-student-publication`, `test`, `translation`, `website`, `working-paper`.

You raise a valid point, if the submission is a preprint rather than a published journal article, `preprint` would be more accurate. 

We could either hardcode `journal-article` since CODECHECK is journal-focused, or make the subject type configurable or derived from the submission type. What would you recommend, should we keep it fixed or add flexibility here?

classes/Orcid/PeerReviewPayloadBuilder.php
subhanLabs: The ORCID peer-review API does not have a `codechecker` role. The available roles are `chair`, `editor`, `member`, `organizer`, `reviewer`, `reviewer-external`, and `senior-editor`. We used `reviewer` as the closest match, Happy to change this if you have a preference.

classes/Orcid/PeerReviewPayloadBuilder.php
subhanLabs: The fallback placeholder values (`Unknown`, `XX`, `CODECHECK Journal`) have been removed. Instead, we validate required journal metadata (publisher name and country) before any deposition attempt. If metadata is missing, the deposit button is disabled and a clear warning is shown pointing editors to the exact settings fields to complete (Journal Settings → Masthead → Publisher / Country).

I also implemented the test deposition button directly in this PR since the settings form was already merged. It is now available in the CODECHECK plugin settings page, it validates journal metadata and makes an authenticated token request to ORCID to verify credentials, without writing any data.

image

classes/Constants.php
subhanLabs: The API type is configurable in the CODECHECK plugin settings page via the "ORCID API" dropdown, which offers two options: "Member Sandbox (development/testing)" and "Member API (production)". Journal managers can switch between them at any time. The sandbox is selected by default to prevent accidental deposits to production profiles during setup and testing.

image

resources/js/Components/CodecheckOrcidSection.vue
subhanLabs: Good call, I have bundled the ORCID iD icon with the plugin as `assets/img/orcid.svg` and updated the component to reference it locally instead of loading it from orcid.org at runtime.

changes.diff
subhanLabs: Good catch, that file was accidentally included. I have removed it.

CodecheckPlugin.php
subhanLabs: Yes, absolutely. The automatic deposition on publication is just a convenience, if the codechecker has not authorised before publication, or if the deposit fails, editors can still trigger it manually at any time using the "Deposit to ORCID" button in the CODECHECK workflow tab. The button remains available after publication so codecheckers can authorise and deposit retroactively.

CodecheckPlugin.php
subhanLabs: I already store the deposition state in the `codecheck_orcid_tokens` table, each row has a `deposit_status` column (`pending`, `success`, `failed`), a `put_code` for successful deposits, and an `error_message` for failures.

The results are surfaced to editors in the CODECHECK workflow tab via the ORCID Deposition section, which shows each codechecker's current status badge (NOT AUTHORISED / PENDING / DEPOSITED / FAILED) and displays the error message inline when a deposit fails. This is loaded via the `orcid-status` API endpoint on every page load so the state is always current.

CodecheckPlugin.php
subhanLabs: Yes, the authorisation was always only possible for the codechecker, the OAuth flow requires them to log in with their own ORCID account, so an editor could never complete it even if they clicked the button. We have now made this explicit in the UI: the authorise button is only shown in the reviewer form, and the editorial workflow view shows a clear note that authorisation must be completed by the codechecker. See the screenshots below showing both views.

Editor:
image

Codechecker:
image

nuest:
@subhanLabs For the sake of documentation, can you please update local docs or the README file which field mentioned in https://support.orcid.org/hc/en-us/articles/360006971333-Peer-Reviews your code sets, and what values are put in there? Thanks!

nuest:
@subhanLabs For the sake of documentation, please post a screenshot of the ORCID Deposition section here.

api/v1/CodecheckApiHandler.php
```diff
@@ -19,10 +19,16 @@
 use APP\plugins\generic\codecheck\classes\CodecheckRegister\CodecheckVenue;
 use APP\plugins\generic\codecheck\classes\Workflow\CodecheckMetadataHandler;
 use APP\plugins\generic\codecheck\classes\Workflow\CodecheckYamlValidator;
+<<<<<<< HEAD
```
nuest: Attention - you committed some merge artefacts here!

classes/Orcid/OrcidAuthHandler.php
nuest: Please localise localize all error messages.

subhanLabs:
Please find screenshots of the ORCID Deposition section attached, the first shows the reviewer form view where the codechecker can authorise and re-deposit, and the second shows the editorial workflow view where the editor can monitor status and trigger deposition.

image

image

subhanLabs:
> @subhanLabs For the sake of documentation, can you please update local docs or the README file which field mentioned in https://support.orcid.org/hc/en-us/articles/360006971333-Peer-Reviews your code sets, and what values are put in there? Thanks!

I have added a section to the README documenting which ORCID peer-review fields are set and what values are used: [Orcid Peer Review Deposition](https://github.com/codecheckers/ojs-codecheck/blob/feature/orcid-deposit/README.md#orcid-peer-review-deposition)

classes/Orcid/PeerReviewPayloadBuilder.php
nuest: @subhanLabs Please revisit the current OJS ORCID integration which type is used there, and if it can be configured? I think it cannot be changed and is hard-coded to `journal-article`

Please also check https://github.com/pkp/pkp-docs/issues/1303 - do we need to set "City" as well?

nuest:
@subhanLabs please merge main branch, it just underwent some massive changes. When done, double check there are no merge artefacts in the code, please.

templates/settings.tpl
```diff
@@ -69,6 +69,41 @@
...
+						.text('✓ ' + data.message)
+						.show();
+				} else {
+					$result
+						.addClass('orcid-test--error')
+						.text('✗ ' + data.error)
+						.show();
+				}
+			})
+			.catch(() => {
```
nuest: Can you catch the message of the error here and show it in the result error below?

classes/Orcid/OrcidDepositService.php
```diff
@@ -0,0 +1,182 @@
...
+                        ?? Constants::ORCID_API_TYPE_SANDBOX;
+
+        if (empty($clientId) || empty($clientSecret)) {
+            CodecheckLogger::error('ORCID deposit skipped: no credentials configured.');
+            return [];
+        }
+
+        $submission = Repo::submission()->get($submissionId);
+        if (!$submission) {
+            return [];
```
nuest: How can this state be reached? Should there not at least be a warning log?

classes/Orcid/OrcidAuthHandler.php
```diff
@@ -0,0 +1,230 @@
...
+        );
+    }
+
+    private function buildRedirectUri($request): string
+    {
+        return $request->getBaseUrl()
+            . '/index.php/index/codecheck/orcid/callback';
+    }
+
+    private function sendPopupError(string $message): void
```
nuest, line 210: Which options in the OJS codebase to communicate an error to the user did you find as an alternative to this approach? How does this pop-up error look like in the UI?

classes/Orcid/OrcidDepositService.php
nuest: The guards should communicate useful information up to the UI for the user to see. Can you create something like an error-result?

classes/Orcid/PeerReviewPayloadBuilder.php
subhanLabs: **`subject-type`** is hard-coded to `journal-article` in our `PeerReviewPayloadBuilder.php`. This mirrors the existing OJS ORCID integration which also hard-codes it, a CODECHECK is always a review of a journal article in this context, so it cannot and should not be configurable.

**City**, according to [pkp/pkp-docs#1303](https://github.com/pkp/pkp-docs/issues/1303), OJS 3.5 added City as a configurable journal-level setting specifically because it is required to submit review contributions via the Member API. However OJS does not expose a publisher city field in the journal context object, so we cannot read it. We have two options:

1. Add a **City** field to our plugin settings (alongside Client ID/Secret) so journal managers can enter it
2. Leave it omitted, ORCID accepts deposits without city in some cases

We went with option 2 for now since OJS itself has no city field. We can add a dedicated city setting in the plugin if ORCID rejects deposits without it.

classes/Orcid/OrcidDepositService.php
```diff
@@ -0,0 +1,182 @@
...
+                return ['orcidId' => $orcidId, 'status' => 'success', 'putCode' => $putCode, 'action' => 'updated'];
+            } else {
+                $newPutCode = $client->postPeerReview($orcidId, $accessToken, $payload);
+                $this->tokenDAO->markSuccess((int) $row->id, $newPutCode);
+                return ['orcidId' => $orcidId, 'status' => 'success', 'putCode' => $newPutCode, 'action' => 'created'];
+            }
+        } catch (\Throwable $e) {
+            $error = $e->getMessage();
+            CodecheckLogger::error('ORCID deposit failed for ' . $orcidId . ': ' . $error);
+            $this->tokenDAO->markFailed((int) $row->id, $error);
```
nuest: Multiple times converting row IDs with `(int)` - maybe this conversion can be flxibly handled within `markSuccess(...)` ?

classes/Orcid/PeerReviewPayloadBuilder.php
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        $org = [
+            'name'    => $publisherName,
+            'address' => [
+                'country' => $country,
+            ],
+        ];
+
+        $city = !empty($journal['publisherCity']) ? trim($journal['publisherCity']) : '';
```
nuest, line 152: The ORCID plugin handles this as an axtra field - I think we should, too.

nuest (changes requested):
See inline comments.

classes/Orcid/OrcidAuthHandler.php
subhanLabs: I looked at OJS notification patterns (`NotificationManager::createTrivialNotification`, JSONMessage, session flash messages) but since the OAuth flow runs in a popup window, standard OJS notifications cannot be used, they would appear in the popup which closes immediately. The postMessage approach is the standard pattern for OAuth popups: the popup sends the error back to the opener window via window.postMessage, closes itself, and the main window displays the error in the ORCID section banner.

Here is a screenshot showing the error as it appears in the main window after the popup closes:


image

classes/Orcid/OrcidDepositService.php
subhanLabs: Done, the deposit method now captures errors from the results array and displays them in the UI. Here is a screenshot showing the error banner:

image

classes/Orcid/OrcidDepositService.php
subhanLabs: Right! the int type hint in the `OrcidTokenDAO.php` already handles the conversion, so the explicit (int) casts in the caller have been removed.

`$this->tokenDAO->markSuccess($row->id, $newPutCode);`

classes/Orcid/PeerReviewPayloadBuilder.php
subhanLabs: Done, added a dedicated City field to the plugin settings following the same approach as the OJS ORCID plugin. The city is now included in the convening-organization address payload (confirmed in logs). ORCID displays "Unknown" for the city when the organization has no disambiguation identifier (Ringgold/ISNI/ROR) — this is expected ORCID behavior for unregistered organizations.

image

classes/Orcid/OrcidDepositService.php
subhanLabs: Good point, added a warning log for this case. In practice this state should not be reachable since deposit is always triggered from a valid submission context, but the log will help diagnose if it ever occurs unexpectedly.

templates/settings.tpl
subhanLabs: Done!
image

classes/Orcid/OrcidAuthHandler.php
subhanLabs: Done!

nuest:
@subhanLabs Please resolve the minor merge conflicts and add a reply to _all_ conversations of https://github.com/codecheckers/ojs-codecheck/pull/126#pullrequestreview-4591333846 - Thanks!

api/v1/CodecheckApiHandler.php
subhanLabs: Resolved!

subhanLabs:
> @subhanLabs Please resolve the minor merge conflicts and add a reply to _all_ conversations of [#126 (review)](https://github.com/codecheckers/ojs-codecheck/pull/126#pullrequestreview-4591333846) - Thanks!

All merge conflicts have been resolved and all review comments have been addressed. Please take another look when you get a chance, thanks!

nuest:
@subhanLabs Please resolve all conversations with a short comment - thanks!

nuest (commented):
pending comments lurking?

nuest:
@subhanLabs — I've merged `main` into this branch and pushed it, so the conflicts
are resolved and CI is green again. `main` had moved 53 commits since your last
merge, including a refactor of the API handler, so I did the integration here
rather than asking you to work through it a third time.

Where the two sides disagreed I kept `main`'s shape, because in each case the
branch was still using something `main` had replaced:

- `api/v1/CodecheckApiHandler.php` — the request cycle moved out of the
  constructor into `execute()`, and every response now goes through
  `respond()`/`JsonResponseEmitter`. Your three ORCID endpoints are transplanted
  onto that; the eleven `JsonResponse::staticResponse()` / `$this->response->response()`
  calls became `respond()`. One thing to be aware of: `respond()` is typed
  `never` and the test emitter *throws* where the production one exits, so I
  moved the success response in `depositToOrcid()` out of its `try` — inside it,
  a success would have been caught and reported as a 500.
- `resources/js/main.js` — the review panel read `publication.codeRepository` and
  `dataRepository`, which #165 removed from the store.
- `classes/Constants.php` — `CODECHECK_ENABLED` was renamed to
  `showArticleSidebar` on `main`.
- `templates/settings.tpl` — your reset button sat where `main` now renders the
  config-version selector; `main` already carries that button further down. Your
  ORCID group is inside `{fbvFormArea}` so it gets the same bordered box as the
  other groups.
- `locale/en/locale.po` — rebuilt from both sides rather than hand-merged:
  `main`'s wording wins for keys both sides had, your 64 new keys are kept, and
  11 keys for settings `main` removed are dropped.

I also added the `CHANGELOG.md` entries, which the branch didn't have.

## One change I made to your code

`OrcidApiClient::httpPost()` and `httpPut()` set `CURLOPT_SSL_VERIFYPEER` and
`CURLOPT_SSL_VERIFYHOST` to `false`. Those are the requests that send the
journal's ORCID client secret to `/oauth/token` and each codechecker's access
token as an `Authorization` header, so certificate verification needs to stay on
— I've removed both options (cURL's defaults are right). I also dropped
`CURLOPT_FOLLOWLOCATION` from `httpPost`: it would replay the `Authorization`
header to whatever host a redirect named, and `postPeerReview()` reads the
put-code out of the `Location` header itself, so following redirects was working
against that anyway. `api/v1/CurlApiClient.php` already sets
`CURLOPT_SSL_VERIFYPEER => true`, so this was the only place it was off.

## Things still open to look at

1. **`onPublicationPublish()`** dereferences
   `Application::get()->getRequest()->getContext()` with no null guard, which is
   the same trap your own commit 7417229 fixed in `callbackTemplateManagerDisplay`.
   It also re-fetches the submission with `Repo::submission()->get()` rather than
   taking it from `$args[2]` — publishing goes through the REST API, where that
   is the reliable source, and `depositToRegister()` right above it does it that
   way. And it logs `$result['orcidId']` for the two failure results that
   `OrcidDepositService` builds without that key, so a journal that enables ORCID
   before entering a client ID gets a PHP warning on every publish.

2. **`depositToOrcid()` ignores the `orcidId` the UI sends.**
   `CodecheckOrcidSection.vue` posts `{submissionId, orcidId}`, but the endpoint
   reads only `submissionId` and calls `depositForSubmission()`, which iterates
   every authorised codechecker. So the per-row "Deposit" button and "Deposit to
   all" do the same thing — with two codecheckers, a targeted re-deposit re-PUTs
   the other person's item too.

3. **`PeerReviewPayloadBuilder` reads `pub-id::doi`**, which isn't on the OJS 3.5
   publication schema (the 3.4 upgrade deletes those rows), so `$articleDoi` is
   always null and no deposited item ever links back to the article. Elsewhere
   the plugin uses `getDoi()` — e.g. `CodecheckMetadataHandler.php:77`.

4. **Checks without a certificate all deposit under the literal external id
   `codecheck:unknown`.** ORCID de-duplicates peer-review items by external id,
   so the second such deposit for the same codechecker returns 409,
   `postPeerReview()` regexes the existing put-code out of the error and returns
   it as if it had created the record, and the UI shows "Deposited". I think
   those deposits should be skipped rather than reported as successful — but
   that's your call on what the right behaviour is.

5. **"Test ORCID Setup" requests the `/group-id-record/update` scope**, not the
   `/activities/update` scope the feature actually uses. A Member client
   provisioned for peer-review deposit but without group-id privileges will be
   reported as broken when it would deposit fine. `depositForSubmission()` also
   calls `createGroupId()` on every deposit and swallows the result into a debug
   log.

6. **The 13 ORCID auth locale keys aren't referenced.** Commit 7beb495 ("localise
   all error messages in OrcidAuthHandler") added them to `locale.po` but changed
   no PHP, so `OrcidAuthHandler` still emits English literals and translators
   would carry 13 dead entries. Related: `CodecheckApiHandler.php` builds one
   message by concatenating a key with `$e->getMessage()`; CLAUDE.md asks for one
   sentence per key with a `{$name}` placeholder, and
   `orcid.auth.error.tokenExchange` already exists in that form.

7. **Smaller things.** The ORCID API-type control is a `<select>`; the settings
   form doesn't use dropdowns any more — multiple-choice settings use
   `.codecheck-choice-list` (see the badge and config-version groups). The
   reviewer-form `MutationObserver` is registered on every backend page and never
   disconnected, though the code it guards returns immediately unless
   `window.codecheckReviewerData` is set. And in `CodecheckOrcidSection.vue`,
   `loadTokenStatus()` has no `else` on `if (response.ok)`, so a failed request
   renders as "No codecheckers have been assigned yet"; `depositAll()` never
   checks `data.results` for failures the way `deposit()` does.

There are also two remaining **conversations** on the earlier review —
the "fallback information" question on `PeerReviewPayloadBuilder` and how an
editor switches between the sandbox and production endpoints? They're marked
outdated but not answered, and I'd like to close them out before merging.

nuest (approved):
Merging now to develop further as part of main codebase, to include feature in upcoming refactorings etc.

PR #134, opened 2026-06-22
Add private checkbox to check repositories

### LINKED ISSUE #100: Add a "Mark as private" checkbox to resources related to the check
We try to accommodate both workflows that are close to an open peer review and ones more classic anonymous.

That is why not all Journals may want to share, or force their authors to share, material related to their work.

We can add a "is private/do not share" checkbox next to the links to code and data repositories that are added so that authors can say: "this is fine for review, but not for resharing".

If that checkbox is ticked, the information should not be published in a related GitHub issue nor added to the codecheck.yml file.
---

## Related User Stories

- C-CO1 (related): As a codechecker, I want the plugin to present the appropriate communication channel...

subhanLabs:
closes #100 

Add `Do not share publicly` checkbox per repository entry in the CODECHECK metadata form. Private repositories are stored with an `isPrivate` flag and filtered from the generated `codecheck.yml`. Backward compatible with existing comma-separated repository format.

image

Repository excluded from the yml preview & download:
image

Private excluded from the article landing page:
image


nuest:
@subhanLabs please check exposure of data to the landing page (and other exports), then remove the "draft" status again.

resources/js/Components/CodecheckMetadataForm.vue
```diff
@@ -534,7 +538,16 @@ export default {
...
           if (data.codecheck.repository) {
-            this.repositories = data.codecheck.repository.split(',').map(r => r.trim()).filter(r => r);
+            try {
+              const parsed = JSON.parse(data.codecheck.repository);
+              // New format: array of {url, isPrivate} objects
+              this.repositories = Array.isArray(parsed)
+                ? parsed
+                : data.codecheck.repository.split(',').map(r => ({ url: r.trim(), isPrivate: false })).filter(r => r.url);
+            } catch {
+              // Backward compat: old plain comma-separated string
```
nuest: For what is this backward compatibility important? We do not have a release of the software yet, so no existing installations do have the "old" data stored.

locale/en/locale.po
```diff
@@ -581,4 +581,7 @@ msgid "plugins.generic.codecheck.status.verificationInProgress"
 msgstr "Code verification is currently in progress."
 
 msgid "plugins.generic.codecheck.additionalContent.frontendLabel"
-msgstr "Additional information"
\ No newline at end of file
+msgstr "Additional information"
+
+msgid "plugins.generic.codecheck.repository.markAsPrivate"
```
nuest, line 844: Add another text with more details, which could be shown hovering over an "(i)" info icon.

Who can see a repository marked as private?

Where is the information _not_ exposed (i.e., not on the landing page, not in the any management issue on GitHub, ...) ?

nuest (changes requested):
- Please add a couple of screenshots how this looks in the UI.
- Please add multiple test cases that ensures correct access to the information in any form or API endpoint or export.

This is a seemingly small feature, but it is one that can really break the trust in the tool for authors and editors if it is not perfect.

nuest:
@subhanLabs Same here - let's resolve the merge conflicts before the review.

This also looks like you have some boilerplae code going for the addtional check box about the source of the codecheck.yml, which is good!

resources/js/Components/CodecheckMetadataForm.vue
subhanLabs: Good point, removed the backward compat code in the refactor commit.

locale/en/locale.po
subhanLabs: Added a tooltip text via the `title` attribute on the label that explains exactly this, the repository will not be shown on the article landing page, not included in the codecheck.yml file, and not posted to the GitHub register issue. Also renamed the label to "Keep private" to be more concise while the tooltip provides the full context on hover.

subhanLabs:
> * Please add a couple of screenshots how this looks in the UI.
> * Please add multiple test cases that ensures correct access to the information in any form or API endpoint or export.
> 
> This is a seemingly small feature, but it is one that can really break the trust in the tool for authors and editors if it is not perfect.

Screenshots added above showing the checkbox in the UI with the tooltip visible on hover.

For test cases: added three PHPUnit unit tests for the `buildYaml()` method covering the three meaningful cases, private repos filtered from YAML, all-private repos resulting in no repository key in YAML at all, and all-public repos all included. Also added three Cypress component tests verifying the checkbox default state, checking/unchecking behaviour, and that the private flag is independent per repository row.

Regarding API/export exposure, checked `CodecheckGithubRegisterApiClient::addIssue()` and confirmed repository URLs are not included in the GitHub register issue body at all (only journal name and submission ID are posted there), so no additional filtering was needed on that side.

classes/Submission/CodecheckSubmissionDAO.php
```diff
@@ -95,7 +95,24 @@ public function getManifest(): array
...
     { 
-        return $this->data['repository'] ?? ''; 
+        $raw = $this->data['repository'] ?? '';
+        if (empty($raw)) {
+            return '';
+        }
+
+        $decoded = json_decode($raw, true);
+
+        if (is_array($decoded) && isset($decoded['repositories']) && is_array($decoded['repositories'])) {
```
nuest, line 106: What if the structure here is not as expected?

At least add an else clause with an error/warning log that something is not working as expected.

classes/Workflow/CodecheckMetadataHandler.php
```diff
@@ -240,9 +240,18 @@ public function buildYaml($publication, $metadata): string
...
-        // Repository
-        if ($repository->repositories) {
-            $data['repository'] = $repository->repositories;
+        // Repository — filter out entries marked as private
+        if ($repository && isset($repository->repositories) && is_array($repository->repositories)) {
+            $publicUrls = array_values(array_map(
+                fn($r) => isset($r->url) ? $r->url : '',
+                array_filter($repository->repositories, fn($r) => empty($r->isPrivate))
+            ));
+            $publicUrls = array_filter($publicUrls);
```
nuest, line 249: Add a debug log statement here to the effect of "filtered out x of n repositories because they are marked as private".

classes/Submission/CodecheckSubmissionDAO.php
```diff
@@ -95,7 +95,24 @@ public function getManifest(): array
 
     public function getRepository(): string 
     { 
-        return $this->data['repository'] ?? ''; 
+        $raw = $this->data['repository'] ?? '';
```
nuest, line 99: Suggest to rename item to `repositories` for clarity, and to not return an empty string but to return an empty array as the "no values" option.

cypress/tests/component/CodecheckMetadataForm.cy.js
```diff
@@ -470,4 +470,69 @@ describe('CodecheckMetadataForm Component', () => {
...
+      .parent()
+      .find('.btn-add')
+      .click();
+
+    // Check only the first one
+    cy.get('.repo-private-checkbox').eq(0).check();
+
+    cy.get('.repo-private-checkbox').eq(0).should('be.checked');
+    cy.get('.repo-private-checkbox').eq(1).should('not.be.checked');
+  });
```
nuest, line 537: Add one more test here: click save, and then the state should be properly preserved. Ideally do not match bit index number, but by the value entered in the field.

nuest (changes requested):

classes/Workflow/CodecheckMetadataHandler.php
subhanLabs: Done, added a debug log statement that logs how many repositories were filtered out.

classes/Submission/CodecheckSubmissionDAO.php
subhanLabs: Done, renamed to getRepositories() returning an array, and added an else clause with a warning log for unexpected format.

cypress/tests/component/CodecheckMetadataForm.cy.js
subhanLabs: Done, added a test that verifies the checkbox state is preserved after save, matching by the URL value entered.

classes/Submission/CodecheckSubmissionDAO.php
subhanLabs: Covered in the `getRepositories()` fix below.

nuest (approved):

PR #138, opened 2026-06-29
feat: Implement proper schema migration structure (Issue #94)

### LINKED ISSUE #94: Add proper schema migration
See discussion in https://github.com/codecheckers/ojs-codecheck/pull/83#discussion_r2723186402
---

## Related User Stories

subhanLabs:
Closes #94

Implements a proper versioned schema migration structure for the CODECHECK plugin, replacing the previous ad-hoc approach where table creation was split across three disconnected methods called manually from `setEnabled()`.

## Approach

This implementation follows the pattern used by the PKP **PLN plugin** ([pkp/pln](https://github.com/pkp/pln)), which is the only official PKP plugin with a full install + upgrade migration structure. The pattern is compatible with OJS 3.5, confirmed via `pkp-lib/stable-3_5_0/classes/plugins/Plugin.php`.

### New structure
```
classes/migration/
├── install/
│   └── CodecheckSchemaMigration.php   ← all 4 tables, fresh install only
└── upgrade/
    └── I94_AddMissingColumns.php      ← defensive column-level changes
```

### How it works

**Fresh install:** `setEnabled()` calls `getInstallMigration()->up()` which creates all 4 tables (`codecheck_metadata`, `codecheck_orcid_tokens`, `codecheck_issue_labels`, `codecheck_status`) and then runs all upgrade migrations. Since columns already exist in the freshly created table, the upgrade migrations do nothing extra.

**Existing install (upgrade):** `Schema::hasTable()` is true so table creation is skipped, existing data is untouched. Upgrade migrations then run and use `Schema::hasColumn()` guards to add only the columns that are missing from older installs.

**OJS Plugin Gallery install:** `getInstallMigration()` hooks into OJS's own `Installer::postInstall` event automatically via the parent `Plugin` class, no manual intervention needed.

**Adding future columns:** Create a new 
`classes/migration/upgrade/I{issue}_Description.php` with a `hasColumn` guard, then add one line calling it at the bottom of the install migration. 
Both fresh installs and existing installs converge to the same schema state.

### Key fixes compared to before

| | Before | After |
|---|---|---|
| Fresh install | 3 methods called manually in `setEnabled()` | Single entry point via `getInstallMigration()` |
| Existing install with new columns | Silently skipped, column never added | `I94_AddMissingColumns` adds missing columns safely |
| OJS gallery install | Bypassed entirely | Hooks into `Installer::postInstall` automatically |
| `down()` | Dropped tables — risked data loss | Throws `DowngradeNotSupportedException` |
| Reset button | Called `down()` from migration | Explicit drops in `resetSchema()`, separate from migration system |

### Other changes
- Remove `classes/migration/CodecheckSchemaMigration.php` (old file)
- Remove unused `CodecheckSchemaMigration` import from `Manage.php`
- Remove `private $migration` property from `CodecheckPlugin`
- Bump `version.xml` from `0.0.0.0` to `1.0.0.0`

## References
- PLN plugin migration pattern: https://github.com/pkp/pln/tree/main/classes/migration
- `Plugin::getInstallMigration()`: https://github.com/pkp/pkp-lib/blob/stable-3_5_0/classes/plugins/Plugin.php#L236


version.xml
```diff
@@ -3,15 +3,15 @@
...
   * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
   *
   * CODECHECK plugin version information.
   -->
 <version>
 	<application>codecheck</application>
 	<type>plugins.generic</type>
-	<release>0.0.0.0</release>
+	<release>1.0.0.0</release>
 	<date>2025-01-01</date>
```
nuest: No need to change the release version as part of this PR. (Especially not without changing the date below it.)

classes/migration/upgrade/I94_AddMissingColumns.php
```diff
@@ -0,0 +1,44 @@
...
+namespace APP\plugins\generic\codecheck\classes\migration\upgrade;
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use PKP\install\DowngradeNotSupportedException;
+
+class I94_AddMissingColumns extends Migration
+{
+    public function up(): void
```
nuest: Add some clever mechanism that ensures the execution of the Migration is nicely logged. Does the Illuminate `Migration` class ensure this?

Maybe something like our own `CodecheckMigration` _abstract_ base class that has an instantiation function that prints its own name? How do PHP classes work, can we all "super.up()" from the up function at start and end?

Please share an installation log output of a fresh installation, which should run the original migration an all upgrades.

It may be easier to solve this via convention/documentation, i.e., just have a log statement at the beginning and end of the `up()` function.

What is your opinon?

nuest (changes requested):
Thanks for the first iteration. Can you please update the documentation as well?

What I find especially important: Where is the _order_ of the upgrade scripts configured?

If I understand correctly, that would be in the `up()` function.

Also make clear in the documentation that "adding a column" is only one thing that could happen, an upgrade script may also process all records (e.g., rename an allowed value), insert content, e.g., for settings, or remove fields/columns/tables that are not used anymore.
_Correct?_

version.xml
subhanLabs: Reverted!

subhanLabs:
> Thanks for the first iteration. Can you please update the documentation as well?
> 
> What I find especially important: Where is the _order_ of the upgrade scripts configured?
> 
> If I understand correctly, that would be in the `up()` function.
> 
> Also make clear in the documentation that "adding a column" is only one thing that could happen, an upgrade script may also process all records (e.g., rename an allowed value), insert content, e.g., for settings, or remove fields/columns/tables that are not used anymore. _Correct?_

Added a docblock to the install migration explaining that the order of calls at the bottom of `runUp()` defines execution order, and that upgrade scripts aren't limited to adding columns, they can rename, migrate data, insert defaults, or remove things too.

classes/migration/upgrade/I94_AddMissingColumns.php
subhanLabs: Added an abstract `CodecheckMigration` base class that wraps `runUp()` with start/end log statements. Each migration now extends this instead of Laravel's `Migration` directly, so logging is guaranteed for all current and future migrations without any extra effort.

nuest (approved):

PR #139, opened 2026-06-29
Feature/publication validation

### LINKED ISSUE #12: Before publication (during copyediting) validate (or allow validation) of CODECHECK metadata
- check that the certificate actually exists under the given DOI
- check that codechecker (ORCID) matches what is published with the certificate (on Zenodo/OSF)
- check that the certificate includes the title of the paper (how to solve cross-referencing issue with DOIs?)

--------

**User stories:**

- _As the codechecker, I want the plugin to help me with creation and storing of the metadata, so that I can focus on the check and have more time for the interesting work._
- _As a publisher, I want the plugin to connect with the CODECHECK infrastructure for publication of certificates and metadata without my staff/my editors having to do work on other tools/platforms, so that it is guaranteed that the CODECHECK follows the CODECHECK principles and requirements._

### LINKED ISSUE #122: Block Publication / Add Warning, when YAML file is structurally invalid
Related to #66

When the generated `codecheck.yml` file is structurally invalid, the publishing of a submission should be blocked / a warning should be sent (similar to #32).

### LINKED ISSUE #143: Make extended publication validation configurable in the plugins settings
options to turn off "hard" pulbication validation for:
- [x] validation of `codecheck.yml` from external repository (#12)
- [x] disallowing pulbication when opted-in CODECHECK wasn't started
- [x] if turned off: silently drop the discussed validation cases

### LINKED ISSUE #144: Show extended publication validation as warnings next to repositories field in CODECHECK metadata form
- [x] use the existing `CodecheckPublicationValidator.php` to validate the metadata form with the `codecheck.yml` file from another repository
- [x] show a warning, if the validation failed

### LINKED ISSUE #145: Support Zenodo & OSF DOI Links for metadata retrieval / import to the metadata form fron an external repositories `codecheck.yml`
See #75

- [x] Zenodo import is already possible, so now import from Zenodo via a DOI should be possible as well
- [x] OSF import is already possible, so now import from OSF via a DOI should be possible as well

### LINKED ISSUE #146: make checkbox for the one repository where codecheck.yml is located
- [x] all repositories should have a radio checkbox to select where the `codecheck.yml` is located
- [x] if the checkbox is checked, all other checkboxes should be uncheckable and the `Load Metadata` Button should only appear for this repository
- [x] if the repository has the 'codecheck.yml', this should be saved in the OJS database as well

dxL1nus:
Closes #12,
closes #122,
closes #143,
closes #144,
closes #145,
closes #146

**!!! Firstly please look at the issue listed above, to see exactly which features this PR introduces.**

- - -

This PR introduces the `CodecheckPublicationValidator.php` which validates things like the CODECHECK status, the produced `codecheck.yml` file and more.

- - -

I reworked the Settings Form UI a bit as well in this PR, to group together related settings (e.g. Settings concerning the GitHub Register, or settings concerning the Submission stage, or settings concerning the publication validation stage).

grafik


CodecheckPlugin.php
```diff
@@ -86,32 +87,19 @@ public function register($category, $path, $mainContextId = null): bool
         return $success;
     }
 
-    public function validateCodecheckStatus(string $hookName, array $args): bool
+    public function validatePublication(string $hookName, array $args): bool
```
nuest: This validation function seems never to return `true` anymore.

nuest (changes requested):
I have just one critical point about a validation function, maybe it must be renamed.

Minor things re. this PR added to a follow-up issue for later, to not hold this merging up: https://github.com/codecheckers/ojs-codecheck/issues/148

CodecheckPlugin.php
dxL1nus: Ohhh now I see what you mean. No that is actually down to something completely different and intended.
It is down to how OJS handles hooks. The publication will be invalid, whenever we ship at least one error in the `$errors` Array. And it will be valid, whenever the array is empty.

The return value of this function doesn't have to do anything with the publication validation. Instead it has to do with how OJS handles this hook. If `true` where to be returned, the Hook: `'Publication::validatePublish'` would stop to be called (meaning that for other plugins & OJS itsself this Hook wouldn't be fired anymore, if we have invalid metadata). This of course means, that e.g. if a submission is in the review stage, but somehow the editor already wants to publish it, we never get to see this error during the publication validation (because our invalid CODECHECK data would stop the hook).

To prevent this, it is best practise to always return `false` on a Hook, so other Plugins and OJS itsself get to call the Hook as well after our plugin did.

nuest: Please document this in the function documentation then (a one line comment next to the "return false" also makes sense), and rename the function also, maybe end it in `...Hook` ?

Then it is clear that the return value is not about the validation result but about the processing of the hook;

dxL1nus: How it looks now:

```php
/**
 * The publication will be invalid, whenever we ship at least one error in the `$errors` Array. And it will be valid, whenever the array is empty.
 * The return value of this function doesn't have to do anything with the publication validation. Instead it has to do with how OJS handles this hook. If `true` where to be returned, the Hook: `'Publication::validatePublish'` would stop to be called (meaning that for other plugins & OJS itsself this Hook wouldn't be fired anymore, if we have invalid metadata). This of course means, that e.g. if a submission is in the review stage, but somehow the editor already wants to publish it, we never get to see this error during the publication validation (because our invalid CODECHECK data would stop the hook).
 * To prevent this, it is best practise to always return `false` on a Hook, so other Plugins and OJS itsself get to call the Hook as well after our plugin did.
 * 
 * @param string $hookName The name of the Hook (`'Publication::validatePublish'`)
 * @param array $args The arguments of the Hook including the validation `$errors` Array at `&$args[0]`
 * 
 * @return bool Returns `false` to enable OJS itsself and other Plugins to continue with their implementation for this Hook
*/
public function validatePublicationHook(string $hookName, array $args): bool
{
    CodecheckLogger::debug("Validating Publication!");
    $errors = &$args[0];
    $codecheckPublicationValidator = new CodecheckPublicationValidator($this);

    $validationErrors = $codecheckPublicationValidator->validatePublication();

    if(is_array($validationErrors)) {
        $errors = array_merge($errors, $validationErrors);
    }

    /* Returns `false` to enable OJS itsself and other Plugins to continue with their implementation for this Hook */
    return false;
}
```

nuest (approved):

